Issue #36, opened 2023-05-18
T1 topology with SONiC-VPP VM image: sonic_vpp_cfg.sh should inject non-contagious numbered interfaces


nikunjdh:
Current implementation of sonic_vpp_cfg.sh script capable of injecting contagious numbered Ethernet interfaces, Ethernet0, Ethernet1, Ethernet3, Ethernet4...

T1 topology bring up requires non-contagious numbered interfaces, interfaces with increment of 4. See below example:

admin@vlab-03:~$ show ip int
Interface    Master    IPv4 address/mask    Admin/Oper    BGP Neighbor    Neighbor IP
-----------  --------  -------------------  ------------  --------------  -------------
Ethernet0              10.0.0.0/31          up/up         ARISTA01T2      10.0.0.1
Ethernet4              10.0.0.2/31          up/up         ARISTA02T2      10.0.0.3
Ethernet8              10.0.0.4/31          up/up         ARISTA03T2      10.0.0.5
Ethernet12             10.0.0.6/31          up/up         ARISTA04T2      10.0.0.7
Ethernet16             10.0.0.8/31          up/up         ARISTA05T2      10.0.0.9
Ethernet20             10.0.0.10/31         up/up         ARISTA06T2      10.0.0.11
Ethernet24             10.0.0.12/31         up/up         ARISTA07T2      10.0.0.13
Ethernet28             10.0.0.14/31         up/up         ARISTA08T2      10.0.0.15
Ethernet32             10.0.0.16/31         up/up         ARISTA09T2      10.0.0.17
Ethernet36             10.0.0.18/31         up/up         ARISTA10T2      10.0.0.19
Ethernet40             10.0.0.20/31         up/up         ARISTA11T2      10.0.0.21
Ethernet44             10.0.0.22/31         up/up         ARISTA12T2      10.0.0.23
Ethernet48             10.0.0.24/31         up/up         ARISTA13T2      10.0.0.25
Ethernet52             10.0.0.26/31         up/up         ARISTA14T2      10.0.0.27
Ethernet56             10.0.0.28/31         up/up         ARISTA15T2      10.0.0.29
Ethernet60             10.0.0.30/31         up/up         ARISTA16T2      10.0.0.31
Ethernet64             10.0.0.32/31         up/up         ARISTA01T0      10.0.0.33
Ethernet68             10.0.0.34/31         up/up         ARISTA02T0      10.0.0.35
[...]

PR #38, opened 2023-05-24
Fix Issue #36 : sonic_vpp_cfg.sh should inject non-contiguous numbere…

patilshashidhar:
…d interfaces

Changed the sonic_vpp_cfg.sh to pass an env variable to generate non-contiguous port names.


patilshashidhar:
Please pass LANES=4 env variable to sonic_vpp_cfg.sh command to generate the required port names

sudo LANES=4 sonic_vpp_cfg -p -i linux_interface_names_comma_seperated

Issue #40, opened 2023-05-31
IP route multi path based on ECMP


patilshashidhar:
SONiC supports ORDERED and UNORDERED ECMP load balancing for routes. The multiple paths are passed as a nexthop group with two or more path members.

https://github.com/sonic-net/SONiC/blob/master/doc/ecmp/ordered_ecmp_next_hop_hld.md

This issue is for multi path routing provisioning on VPP.

PR #41, opened 2023-05-31
Fix Issue #40 : IP route multi path based on ECMP

patilshashidhar:
The NEXTHOP GROUP configuration is done in Make-Before-Break way by SONiC. Currently only (Un)Ordered ECMP supported. The route config comes as an update for a nexthop group.

Unit testing done with 3 nexthops for a given route and verified.


patilshashidhar:
The xcvrd tests failed because of a timing issue. Otherwise the build went through fine.

=========================== short test summary info ============================
FAILED tests/test_xcvrd.py::TestXcvrdScript::test_SfpStateUpdateTask_task_run_stop
=================== 1 failed, 69 passed, 1 warning in 15.84s ===================
[  FAIL LOG END  ] [ target/python-wheels/bullseye/sonic_xcvrd-1.0-py3-none-any.whl ]

PR #60, opened 2023-07-18
Added dual-stack support for loopback interfaces

rameshraghupathy:
Added dual-stack support for loopback interfaces


patilshashidhar:
Please use the commit format I mentioned in the email. You need to enter below text during the commit
Fixes issue #XX - Your issue description
    
Fix details
    
Signed-off-by: Your name <github-registered-emailid>

Please raise another PR with details I asked.

platform/saivpp/vpplib/SwitchStateBaseRif.cpp
```diff
@@ -855,12 +861,57 @@ sai_status_t SwitchStateBase::process_interface_loopback (
...
-        vpp_add_del_lpb_intf_ip_addr(serializedObjectId, is_add);
+        std::string vppIfName = "loop" + std::to_string(getInstanceFromHostIfname(interfaceName));
+
+        if (is_add && (lpbInstMap.find(vppIfName) != lpbInstMap.end()))
+        {
+            // interface already exists - store dual stack IP
+            lpbIpToIfMap[destinationIP] = vppIfName;
+            lpbIpToHostIfMap[destinationIP] = interfaceName;
+            SWSS_LOG_DEBUG("interfaceName:%s exists new-ip:%s",
+                interfaceName.c_str(), destinationIP.c_str());
```
patilshashidhar, line 872: Aren't you supposed to configure the IPv6 address in the VPP ? You are just returning here without adding v6 address. This fix will not work.

platform/saivpp/vpplib/SwitchStateBaseRif.cpp
```diff
@@ -855,12 +861,57 @@ sai_status_t SwitchStateBase::process_interface_loopback (
...
+        {
+            vpp_add_del_lpb_intf_ip_addr(serializedObjectId, is_add);
+        }
     }
 
     return SAI_STATUS_SUCCESS;
 }
 
+void SwitchStateBase::eraseDualStackEntries(const std::string& destinationIP) {
+  // Find the entries in both maps based on the destination IP
```
patilshashidhar, line 883: Not sure why you need this function.

patilshashidhar:
You will need to refactor the current  loopback interface create/address update code.
Separate loopback instance creation/deletion and the IP(v4/v6) address updation.
The logic should be 
1. If a loopback instance for the LoopbackX interface in linux stack does not exist, then create an instance. After creation go step 3.
2. If an instance of loop exists in VPP use the instance for IP configuration in step 3.
3. Figure out the IP address type and configure the address in the VPP for the loop instance found in either step 1 or 2.

For loopback delete the loop instance should be deleted in VPP after all addresses are deleted on the LoopbackX locally.

platform/saivpp/vpplib/SwitchStateBaseRif.cpp
rameshraghupathy: Looks like the address update happens as part of the existing code itself. Anyhow, I added a function to update the address as you suggested.

platform/saivpp/vpplib/SwitchStateBaseRif.cpp
rameshraghupathy: Removed.  Handling  this situation in a better way now.  This was needed for the case of dual stack, where a Loopback interface is having two addresses and the delete happens with one of the address.

rameshraghupathy:
> 
Made all the changes need to meet the following requirements:

1. If a loopback instance for the LoopbackX interface in linux stack does not exist, then create an instance. After creation go step 3.

2. If an instance of loop exists in VPP use the instance for IP configuration in step 3.

3. Figure out the IP address type and configure the address in the VPP for the loop instance found in either step 1 or 2.

For loopback delete the loop instance should be deleted in VPP after all addresses are deleted on the LoopbackX locally.

yue-fred-gao:
The change has been committed.

PR #91, opened 2024-02-28
ISSUE 90: SONiC-VPP LAG feature support

bendrapubalareddy:


platform/saivpp/vpplib/SwitchStateBaseFdb.cpp
```diff
@@ -1082,3 +1110,264 @@ sai_status_t SwitchStateBase::vpp_delete_bvi_interface(
...
+    if (lag_ifname == NULL)
+    {
+        SWSS_LOG_NOTICE("LAG interface name is not found for LAG PORT :%s",sai_serialize_object_id(lag_oid).c_str());
+        return SAI_STATUS_FAILURE;
+    }
+
+    //Delete the Bond interface
+    delete_bond_interface(lag_ifname);
+    remove_lag_to_bond_entry(lag_oid);
+
```
patilshashidhar, line 1215: I think you need to call refresh_interfaces_list() to build new interface id to name mapping once bond interface is deleted.

platform/saivpp/vpplib/SwitchStateBaseFdb.cpp
```diff
@@ -1082,3 +1110,264 @@ sai_status_t SwitchStateBase::vpp_delete_bvi_interface(
...
+    uint32_t swif_idx = ~0;
+    const char *hw_ifname;
+    SWSS_LOG_ENTER();
+
+    //set mode and lb. ONiC config does not have provision to pass mode and load balancing algorithm
+    mode = VPP_BOND_API_MODE_ROUND_ROBIN;
+
+    //if LACP is to be enabled in VPP
+    //mode = VPP_BOND_API_MODE_LACP;
+    lb = VPP_BOND_API_LB_ALGO_L2;
```
patilshashidhar, line 1170: I think the hash algorithm should cover L4 and L4 headers for flow consistency to avoid reordering problem.

platform/saivpp/vpplib/vppxlate/SaiVppXlate.c
```diff
@@ -1011,6 +1092,20 @@ static u32 get_swif_idx (vat_main_t *vam, const char *ifname)
     return ((u32) -1);
 }
 
+static const char * get_swif_name (vat_main_t *vam, const u32 swif_idx)
+{
+    hash_pair_t *p;
```
patilshashidhar, line 1097: Do you think a VPP_LOCK() is required while iterating over the list ?

platform/saivpp/vpplib/SwitchStateBaseFdb.cpp
bendrapubalareddy: Ideally this should be configurable , which SONiC does not support. I am keeping round-robin as default one. Based on the requirement one has to make it configurable which SONiC does not support. May be we will raise an issue to track it.

platform/saivpp/vpplib/SwitchStateBaseFdb.cpp
bendrapubalareddy: Yes, done

platform/saivpp/vpplib/vppxlate/SaiVppXlate.c
bendrapubalareddy: ok

platform/saivpp/vpplib/SwitchStateBaseFdb.cpp
yue-fred-gao: There is no one-size-fit-all solution here. If the LAG is for L2 switching, we should use L23. If it is for L3 routing, L34 is better, which is the typical 5 tuples hashing. But we don't know if a LAG is for L2 or L3 at the time it is created. Maybe we can create lag with a default lb algo like l23 and change algo to l34 if the lag has IP configured? To do that, either we have to delete and recreate lag in vpp or we augment vpp to allow changing lb algo after lag is created.

AkeelAli (approved):
Just needs to be adapted to latest to resolve conflicts.

yue-fred-gao:
@AkeelAli, the conflict can't be resolved using web editor. Can you please update the PR?

AkeelAli:
I think I need access rights to the fork to push a change. @bendrapubalareddy can you please provide me access rights to your fork: https://github.com/bendrapubalareddy/sonic-platform-vpp/tree/main or let me know if there's another way to go about this.

bendrapubalareddy:
> I think I need access rights to the fork to push a change. @bendrapubalareddy can you please provide me access rights to your fork: https://github.com/bendrapubalareddy/sonic-platform-vpp/tree/main or let me know if there's another way to go about this.

@AkeelAli i have resolved the conflicts with latest main. It's ready to merge. Please check and merge. Thanks.

bendrapubalareddy:
@AkeelAli i have resolved the conflicts and it's upto date with latest
main. Please merge now.

PR #95, opened 2024-03-14
Update sonic-buildimage to the latest 202311 build: 4ae9a3d1e1

Giluerre:
**This pull request makes following changes:**

Fixes issue:
- #94 

I have created issue for this solution I am submitting

PR #98, opened 2024-07-03
VxLAN EVPN support

oholubnychyy:
The PR add support for VxLAN tunnel configured by BGP EVPN.

First commit is a cherry-pick from PR #95 just to fix build issue. It shouldn't be a part of this PR but without it pipeline will fail.
Second commit actually the feature.


oholubnychyy:
As soon as no pipeline is triggered, there is no need in commit which fix build issue.
Closing this PR and open new one with VxLAN feature only.

Issue #125, opened 2024-10-24
Optimize the makefile for vppbld


linqingxuan:
Hello everyone, I've been encountering errors while compiling on some machines with docker-in-docker. Later, I realized that docker-in-docker doesn't serve much purpose for us. We can completely remove this part and directly use the Sonic compilation container for building, aligning with other platforms. This will reduce the intermediate build process for constructing the docker image. After testing, it runs without any issues.

PR #127, opened 2024-10-24
Optimize the makefile for vppbld(#125)

linqingxuan:
--Compile VPP deb directly using the Sonic container.

### DIFF
--- a/platform/vpp/vppbld/Makefile
+++ b/platform/vpp/vppbld/Makefile
@@ -93,8 +93,13 @@ docker_build:  vppbld-docker/Dockerfile-$(BLDENV)
 	rm -rf debs
 	popd
 
-$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% : docker_build repo_clone
-	$(VPP_DOCKER_RUN) /run_build.sh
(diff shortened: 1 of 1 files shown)


platform/vpp/vppbld/Makefile
```diff
@@ -93,8 +93,13 @@ docker_build:  vppbld-docker/Dockerfile-$(BLDENV)
 	rm -rf debs
 	popd
 
-$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% : docker_build repo_clone
-	$(VPP_DOCKER_RUN) /run_build.sh
+$(addprefix $(DEST)/, $(MAIN_TARGET)): $(DEST)/% :  repo_clone
+
+	pushd $(VPP_REPO_DIR)
+	make UNATTENDED=y PLATFORM=vpp install-dep
```
yue-fred-gao, line 99: With this change, install-dep will install some packages to slave docker. It may cause build and runtime issues to other components. So I chose to use its own container to build vpp

linqingxuan: I didn't think it through.

PR #133, opened 2024-11-04
Fixed some dbg package naming issues

linqingxuan:
Fix LIBSAIREDIS_DBG naming issues
The reference name of LIBSAIREDIS_DBGSYM in other rule files is LIBSAIREDIS_DBG
image

Fix SYNCD_VPP_DBG naming issues
The reference name of SYNCD_VPP_DBGSYM in other rule files is SYNCD_VPP_DBG
image


platform/mkrules/rules/sairedis.mk
```diff
@@ -48,20 +48,20 @@ LIBSAIMETADATA_DEV = libsaimetadata-dev_$(LIBSAIREDIS_VERSION)_$(CONFIGURED_ARCH
 $(LIBSAIMETADATA_DEV)_DEPENDS += $(LIBSAIMETADATA)
 $(eval $(call add_derived_package,$(LIBSAIREDIS),$(LIBSAIMETADATA_DEV)))
 
-LIBSAIREDIS_DBGSYM = $(LIBSAIREDIS_NAME)-dbgsym_$(LIBSAIREDIS_VERSION)_$(CONFIGURED_ARCH).deb
-$(LIBSAIREDIS_DBGSYM)_DEPENDS += $(LIBSAIREDIS)
-$(LIBSAIREDIS_DBGSYM)_RDEPENDS += $(LIBSAIREDIS)
-$(eval $(call add_derived_package,$(LIBSAIREDIS),$(LIBSAIREDIS_DBGSYM)))
+LIBSAIREDIS_DBG = $(LIBSAIREDIS_NAME)-dbgsym_$(LIBSAIREDIS_VERSION)_$(CONFIGURED_ARCH).deb
```
VladimirKuk, line 51: If we want to make consistent naming across the code, I'd suggest to also rename swss-dbg_1.0.0_amd64.deb to swss-dbgsym_1.0.0_amd64.deb

linqingxuan: Do you mean to add "sym" to the reference? I have considered this, but I have seen that the official designs of different platforms all remove "sym". I suggest that we gradually remove sym in the makefile, so that it will be easier to follow up the community version upgrade later.

linqingxuan:
image

platform/mkrules/rules/sairedis.mk
VladimirKuk: I was under impression that it is the other way. 
I've looked at the rules directory and there most of the packages using dbgsym, and most of the "dbg" usage is by dockers.

linqingxuan: So a better way is to change it to use dbgsym naming, I will submit the modification next time

Issue #138, opened 2024-11-08
test_vxlan_ecmp.py failed occasionally


yue-fred-gao:
test_vxlan_ecmp.py failed occasionally. It appears that vpp_vxlan_tunnel_add_del of a new tunnel sometimes mistakenly returns the sw_ifindex of the vxlan tunnel deleted right before it. The suspicion is that set_reply_status in vl_api_bfd_udp_session_event_t_handler, which is unsolicited event callback, messes up the call sequence.

linqingxuan:
I was testing vxlan-related use cases recently, and I just found it. I am checking the code. If you have code updates, please @ me and I want to learn from it. If I can fix the whole problem, I will also submit a PR. Thank you very much

yue-fred-gao:
@linqingxuan, I have created a PR to fix this issue. With the fix, the sonic mgmt test passed 5 times in a roll. Please take a look.

linqingxuan:
I applied this patch and it did solve the problem. I recently had a requirement to have different vxlan  macs based on the inner vxlan subnet.
For example
![image](https://github.com/user-attachments/assets/317f34c5-767f-4005-be0e-97e964e6c592)
1 2 3 are three VMs, they have different MACs, but they do not accept vxlan arp layer 2 learning, which is quite awkward. I am currently using centec and broadcom chip switches, which can accept routing with MACs. I recently read the vpp source code and it seems that it does not support this scenario. In the tofino chip, I used acl to modify the inner MAC in egress.
If this scenario should be modified in vpp? If you can give me some advice, I will be very grateful

linqingxuan:
用这个可以吗https://docs.fd.io/vpp/17.04/clicmd_src_vnet_l2.html

linqingxuan:
Is it okay to use this?https://docs.fd.io/vpp/17.04/clicmd_src_vnet_l2.html

yue-fred-gao:
Is it inner source Mac or destination Mac that you want to change based on subnet? And again, is it subnet of source IP or destination IP?

linqingxuan:
base inner destination IP change destination Mac

linqingxuan:
![image](https://github.com/user-attachments/assets/adf4156b-168b-4480-b224-c0fff7472662)
I have refined the content of the picture. As shown in the picture, I expect to dynamically change the inner mac of the datagram based on vni and inner ip. They are a vm under bm, and bm has closed vxlan layer 2 learning.

[...]

PR #139, opened 2024-11-15
Issue #138: Fix instability in test_vxlan_ecmp sonic-mgmt test

yue-fred-gao:
### why

Addressing issue #138:
test_vxlan_ecmp sonic-mgmt test failed occasionally. This is because vl_api_bfd_udp_session_event_t_handler calls set_reply_status. The callback is for unsolicited event (bfd state update). set_reply_status should not be called for unsolicited event and it can mess up other requests. In our case, it is vxlan tunnel create request.

In the long run, we should make request/response more robust by setting request_message_id in the context of request to match its response.

### What this PR does
remove set_reply_status from vl_api_bfd_udp_session_event_t_handler
Also fixed bridge domain ID resource leak. BD ID is not put back to the resource pool when vxlan decap is deleted.

### DIFF
--- a/saivpp/src/TunnelManager.cpp
+++ b/saivpp/src/TunnelManager.cpp
@@ -320,6 +320,7 @@ TunnelManager::create_vpp_vxlan_decap(
     vpp_status = create_bvi_interface(bvi_mac, bd_id);
     if (vpp_status != 0) {
         SWSS_LOG_ERROR("Failed to create bvi interface");
+        m_switch_db->dynamic_bd_id_pool.free(bd_id);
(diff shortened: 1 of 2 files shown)


AkeelAli (approved):

kcudnik:
Hey, we have PR https://github.com/sonic-net/sonic-sairedis/pull/1473 that have move src from sonic-platform-vpp to virtual switch in sairedis but its not compiling yet, since we dont have pipeline setup in this repo to produce deb packages and i alreayd saw taht you are adding new changes here which get out of sync + @lguohan

PR #163, opened 2025-02-25
Change router interface MTU to L3 payload

yue-fred-gao:
### why
vpp supports per protocol mtu: L3, ipv4, ipv6 and mpls. However, each time we set mtu for one protocol, it resets mtu of other protocols to 0. For a packet to be forwarded, it checks the mtu for the corresponding protocol. If the mtu is 0, it checks mtu of L3. If L3 is also 0, it uses 9000 as default. (see https://github.com/FDio/vpp/blob/master/src/vnet/interface_funcs.h#L317). This causes sonic-mgmt mtu test failed at 9114, which sets router interface MTU to 9100.
### what this PR does
sonic router interface doesn't support per protocol MTU. We set the mtu to protocol L3 in vpp, which applies to all L3 payloads.


dypet (approved):

dypet:
LGTM

PR #210, opened 2026-02-25
SONiC-VPP trixie migration.

dypet:
Migrating to Trixie, corresponding buildimage change: https://github.com/sonic-net/sonic-buildimage/pull/25677


yue-fred-gao (approved):
I will merge once the build image PR is approved to minimize inconsistencies.

yue-fred-gao:
I found the issue that interfaces didn't come up. In 2602, they moved tap interface functionalities to a separate plugin. You need to add "plugin tap_plugin.so { enable }" to startup.conf. It works with the new plugin added and ping is working.

PR #220, opened 2026-04-13
Increase vpp dpdk tx buffer to handle burst

yue-fred-gao:
### why
in sonic-mgmt test_ip_packet, test cases generate 1000 packets and verify received packets in ptf. In azure sonic-vpp PR checker, we saw intermittent failure due to packet drop with reason "dpdk tx failure". This is because KVM virtio NIC could not drain the packets fast enough. When tx ring buffer is full, packets will be dropped with "dpdk tx failure".

### what this PR does
increase num-tx-desc to accommodate the burst of 1000 packets


dypet (approved):
LGTM

PR #221, opened 2026-04-14
Fix p2mp tunnel

lolyu:
This PR is to adapt VPP P2MP to have the same behavior as SAI P2MP (accept IPinIP packets with any source IP).
Please refer to this HLD for more details: https://github.com/sonic-net/sonic-platform-vpp/pull/222


yue-fred-gao:
@lolyu thanks for the contribution. Can you please create a PR in fd.io? We normally submit the change to fd.io upstream unless it is some change not interested to the community, for example, the tunnel term ACL plugin is committed to sonic-vpp only.

yue-fred-gao:
I saw you already created PR in fd.io. Please follow through with it.

lolyu:
> I saw you already created PR in fd.io. Please follow through with it.

Hi @yue-fred-gao, I don't plan to upstream this change to fd.io, as it is a hacky adaptation to VPP to allow SONiC-favored P2MP tunnel decap behavior, please check the HLD here for more details: https://github.com/sonic-net/sonic-platform-vpp/pull/222, section 8.

PR #222, opened 2026-04-15
Add SONiC VPP IPinIP tunnel HLD

lolyu:
Add IPinIP tunnel encap/decap support to SONiC VPP.

* Microsoft ADO (number only): 37902558

|Repo|PR|state|
|-|-|-|
|sinic-platform-vpp|[Add patch to support mp2p ipip tunnel](https://github.com/sonic-net/sonic-platform-vpp/pull/226)|GitHub issue/pull request detail|
|sonic-sairedis|[[vpp] add ipinip tunnel encap/decap support](https://github.com/sonic-net/sonic-sairedis/pull/1860)|GitHub issue/pull request detail|

### DIFF
--- a/docs/HLD/vpp-ipiptunnel.md
+++ b/docs/HLD/vpp-ipiptunnel.md
@@ -159,11 +159,11 @@ SAI_OBJECT_TYPE_TUNNEL (defines tunnel endpoints + TTL/DSCP modes)
 │
 └── SAI_OBJECT_TYPE_NEXT_HOP (actionable — triggers encap for routes)
     ├── SAI_NEXT_HOP_ATTR_TYPE        = SAI_NEXT_HOP_TYPE_TUNNEL_ENCAP
-    ├── SAI_NEXT_HOP_ATTR_IP          = destination IP (inner packet dst)
+    ├── SAI_NEXT_HOP_ATTR_IP          = tunnel destination IP (peer endpoint, becomes outer dst)
     └── SAI_NEXT_HOP_ATTR_TUNNEL_ID   = reference to tunnel object above
 ```
 
-The **tunnel object** defines the P2P endpoints and QoS modes. For encap (unlike decap), the tunnel object carries `ENCAP_SRC_IP` and `ENCAP_DST_IP` that become the outer header source and destination.
+The **tunnel object** defines the P2P endpoints and QoS modes. For encap, the tunnel object carries `ENCAP_SRC_IP` (our loopback) which becomes the outer header source IP. The **nexthop's** `SAI_NEXT_HOP_ATTR_IP` provides the peer endpoint IP that becomes the outer header destination — this is the same value as the tunnel's `ENCAP_DST_IP` in the MuxOrch DualToR use case, but the VPP implementation reads it from the nexthop object.
 
 The **tunnel encap nexthop** is the actionable object — routes pointing to this nexthop will have their packets encapsulated through the tunnel. `MuxOrch` creates these nexthops for active-standby traffic steering.
 
--- a/docs/HLD/vpp-ipiptunnel.md
+++ b/docs/HLD/vpp-ipiptunnel.md
@@ -1,0 +1,745 @@
+# IPinIP Tunnel on VPP Dataplane — High Level Design
+
+## Table of Contents
+
+1. [Revisions](#1-revisions)
+2. [Scope](#2-scope)
+3. [Overview](#3-overview)
+4. [Feature Description](#4-feature-description)
+5. [SONiC Configuration](#5-sonic-configuration)
+6. [SAI API Calls](#6-sai-api-calls)
+7. [VPP Implementation](#7-vpp-implementation)
+8. [VPP P2MP Tunnel SONiC Adaptation](#8-vpp-p2mp-tunnel-sonic-adaptation)
+9. [Appendix A: File Change Summary](#appendix-a-file-change-summary)
+10. [Appendix B: VPP IPIP API Reference](#appendix-b-vpp-ipip-api-reference)
+
+---
+
+## 1. Revisions
+
+| Rev | Date | Author(s) | Changes |
+|-----|------|-----------|---------|
+|v0.1 | 04/12/2026 | Longxiang Lyu (lolv@microsoft.com) | Initial Draft |
+
+---
+
+## 2. Scope
+
+This document describes the IPinIP (IPv4/IPv6) tunnel support for SONiC VPP.
+
+---
+
+## 3. Overview
+
+| Term | Meaning                          |
+| ---- | -------------------------------- |
+| TEIB | Tunnel Endpoint Information Base |
+
+This HLD covers IPinIP tunnel support for the VPP dataplane in SONiC, including both **decapsulation** (P2MP/P2P) and **encapsulation** (P2P) paths.
+
+- **Decap**: Handles P2MP and P2P decap tunnels triggered by `SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY`, used by `TunnelDecapOrch` for features such as MuxCable/Dualtor.
+- **Encap**: Handles P2P encap tunnels triggered by `SAI_NEXT_HOP_TYPE_TUNNEL_ENCAP` nexthops, used by `MuxOrch` for active-standby DualToR traffic steering.
+
+| Feature | Description | VPP Core Change | SAI VPP Change |
+|---|---|---|---|
+| IPinIP Decap (P2MP) | Decapsulate IPinIP packets matching a local DST IP and any SRC IP | Yes — P2MP lookup fix in `ipip4-input` | Yes |
+| IPinIP Decap (P2P) | Decapsulate IPinIP packets from a specific peer | None — native IPIP | Yes |
+| IPinIP Encap (P2P) | Encapsulate packets and send through IPinIP tunnel | None — native IPIP | Yes |
+
+---
+
+## 4. Feature Description
+
+### 4.1 Decap — How It Works
+
+SONiC creates IPinIP decap tunnels to handle encapsulated traffic arriving at
+the switch. The typical use case is:
+
+1. A remote peer encapsulates a packet with an outer IP header (src=peer, dst=our loopback)
+2. The packet arrives at the switch's underlay interface
+3. SONiC VPP dataplane matches the outer DST IP against a SONiC tunnel decapsulation term entry
+4. SONiC VPP dataplane strips the outer IP header (decapsulation)
+5. The inner packet is forwarded normally via the overlay FIB
+
+```
+Encapsulated packet arrives
+│
+┌─────────────────────────────────────────────────────────┐
+│  SONiC VPP Dataplane                                    │
+│                                                         │
+│  ┌──────────────┐     ┌──────────────────────┐          │
+│  │ ip4-input    │────>│ ipip4-input          │          │
+│  │              │     │ (tunnel lookup)      │          │
+│  └──────────────┘     └──────────┬───────────┘          │
+│                                  │ match: strip outer IP│
+│                                  ▼                      │
+│                       ┌──────────────────────┐          │
+│                       │ ip4-input (inner pkt)│          │
+│                       │ route via overlay FIB│          │
+│                       └──────────┬───────────┘          │
+│                                  ▼                      │
+│                       ┌──────────────────────┐          │
+│                       │ ip4-rewrite → tx     │          │
+│                       └──────────────────────┘          │
+│                                                         │
+└─────────────────────────────────────────────────────────┘
+```
+
+### 4.2 Encap — How It Works
+
+SONiC creates IPinIP encap tunnels to steer traffic through a P2P tunnel to a
+specific peer. The typical use case is MuxCable/Dualtor active-standby, where
+the standby ToR must forward traffic through the IPinIP tunnel to the active ToR.
+
+1. `MuxOrch` (or `TunnelDecapOrch`) creates a P2P tunnel object with `SAI_TUNNEL_ATTR_ENCAP_SRC_IP` and `SAI_TUNNEL_ATTR_ENCAP_DST_IP`
+2. A `TUNNEL_ENCAP` nexthop is created referencing the tunnel + destination IP
+3. Routes are programmed to use this tunnel nexthop
+4. Matching packets are encapsulated with the outer IP header and forwarded
+
+```
+Packet to encap destination
+│
+┌─────────────────────────────────────────────────────────┐
+│  SONiC VPP Dataplane                                    │
+│                                                         │
+│  ┌──────────────┐     ┌──────────────────────┐          │
+│  │ ip4-lookup   │────>│ ipN-midchain         │          │
+│  │ (route hit)  │     │ (tunnel adj)         │          │
+│  └──────────────┘     └──────────┬───────────┘          │
+│                                  │ add outer IP header  │
+│                                  ▼                      │
+│                       ┌──────────────────────┐          │
+│                       │ ipip4-encap          │          │
+│                       │ (src=ENCAP_SRC_IP    │          │
+│                       │  dst=ENCAP_DST_IP)   │          │
+│                       └──────────┬───────────┘          │
+│                                  ▼                      │
+│                       ┌──────────────────────┐          │
+│                       │ adj-midchain-tx      │          │
+│                       │ → ip4-rewrite → tx   │          │
+│                       └──────────────────────┘          │
+│                                                         │
+└─────────────────────────────────────────────────────────┘
+```
+
+### 4.3 SAI Object Model — Decap
+
+Two SAI objects are involved:
+
+```
+SAI_OBJECT_TYPE_TUNNEL (metadata — defines TTL/DSCP/ECN modes)
+│
+└── SAI_OBJECT_TYPE_TUNNEL_TERM_TABLE_ENTRY (actionable — triggers decap)
+    ├── SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_VR_ID             = virtual router for overlay
+    ├── SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TYPE              = P2MP (any source) or P2P (specific peer)
+    ├── SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_DST_IP            = our loopback IP (match outer dst)
+    ├── SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_SRC_IP            = remote peer IP (P2P only)
+    ├── SAI_TUNNEL_TERM_TABLE_ENTRY_ATTR_TUNNEL_TYPE       = SAI_TUNNEL_TYPE_IPINIP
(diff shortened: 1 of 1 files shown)


lolyu (commented):
### ✅ Review — Thorough HLD

Excellent coverage of both decap (P2MP/P2P) and encap (P2P) paths. The SAI↔VPP semantic mismatch analysis (Section 7) is particularly valuable — clearly explains why VPP's TEIB-based allowlist doesn't match SAI's permissive P2MP model, and the adaptation (P2MP wildcard fallback patch) is well-justified.

Minor:
- Section numbering gap: 6.2 → 6.5 (missing 6.3, 6.4)
- Consider noting IPv6 scope (in/out for this phase)
- The P2MP permissive decap security trade-off could use a one-liner about underlay ACL/routing being the security boundary

LGTM 🚀

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+└─────────────────────────────────────────────────────────┘
+```
+
+### 4.2 Encap — How It Works
+
+SONiC creates IPinIP encap tunnels to steer traffic through a P2P tunnel to a
+specific peer. The typical use case is MuxCable/Dualtor active-standby, where
+the standby ToR must forward traffic through the IPinIP tunnel to the active ToR.
+
+1. `MuxOrch` (or `TunnelDecapOrch`) creates a P2P tunnel object with `SAI_TUNNEL_ATTR_ENCAP_SRC_IP` and `SAI_TUNNEL_ATTR_ENCAP_DST_IP`
```
yue-fred-gao, line 94: TunnelDecapOrch does encap job?

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+
+    IPIP->>DB: 1. Read tunnel attrs (ENCAP_SRC_IP, ENCAP_DST_IP, TTL/DSCP/ECN)
+    DB-->>IPIP: tunnel attributes
+    IPIP->>IPIP: map_sai_to_vpp_flags()
+    IPIP->>IPIP: 2. Build vpp_ipip_tunnel_t (src=ENCAP_SRC_IP, dst=ENCAP_DST_IP, P2P)
+
+    IPIP->>VPP: 3. vpp_ipip_tunnel_add(&req)
+    VPP-->>IPIP: sw_if_index (e.g. ipip1)
+
+    IPIP->>VPP: 4. refresh_interfaces_list()
```
yue-fred-gao, line 432: vpp_ipip_tunnel_add already returns sw_if_index. Do we still need refresh_interfaces_list?

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+#### New / Modified Functions
+
+| Function | VPP API Message | Description |
+|---|---|---|
+| `vpp_ipip_tunnel_add(vpp_ipip_tunnel_t *tunnel, uint32_t *sw_if_index)` | `IPIP_ADD_TUNNEL` | Creates an IPIP tunnel interface in VPP. Returns the allocated `sw_if_index` (e.g. `ipip0`). |
+| `vpp_ipip_tunnel_del(uint32_t sw_if_index)` | `IPIP_DEL_TUNNEL` | Deletes an IPIP tunnel interface by its `sw_if_index`. |
+| `sw_interface_set_unnumbered(uint32_t unnumbered_sw_if_index, uint32_t ip_sw_if_index, bool is_add)` | `SW_INTERFACE_SET_UNNUMBERED` | Configures an unnumbered interface: `unnumbered_sw_if_index` borrows its IP from `ip_sw_if_index`. |
+| `vpp_sw_interface_find_by_ip(vpp_ip_addr_t *search_ip, uint32_t *out_sw_if_index)` | `IP_ADDRESS_DUMP` | Finds the `sw_if_index` of the interface owning a given IP address. Iterates all known interfaces from `interface_name_by_sw_index` hash via per-interface `IP_ADDRESS_DUMP` queries. Returns `-ENOENT` if no match. |
+
+#### `vpp_sw_interface_find_by_ip` — Implementation Note
```
yue-fred-gao, line 504: You need add vrf to the match condition. between vrfs, ip can be duplicated.

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+
+The function finds the interface owning the IP address by:
+
+1. Collecting all known `sw_if_index` values from the `interface_name_by_sw_index` clib hash into a `vec`.
+2. Iterating each interface individually, sending `IP_ADDRESS_DUMP` per interface.
+3. The reply handler (`vl_api_ip_address_details_t_handler`) compares the dumped address against the search target; on match it records the `sw_if_index`.
+
+---
```
yue-fred-gao, line 512: Please add ipinip tunnel config in CONFIG_DB/APP_DB. It is also helpful to list vpp binary APIs for ipinip tunnel so we know from top to down how sonic maps to vpp

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+
```
yue-fred-gao, line 513: Is ipinip tunnel in SONiC bidirectional or you config encap and decap separately? If it is bidirectional, or user configures both encap and decap for bidirectional traffic with same endpoints, will there be collision when both encap and decap trying to create ipinip tunnel in vpp with vpp_ipip_tunnel_add?

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,745 @@
...
+    subgraph decap["ipip4-input decap path (with SONiC-VPP patch)"]
+        pkt["Incoming IPinIP packet"]
+        build["Build key: {P2P, src=outer_dst, dst=outer_src}"]
+        lookup["ipip_tunnel_db_find"]
+        hit{"P2P hit?"}
+        ok["Decap → deliver inner pkt to ipip0"]
+
+        subgraph patch["SONiC-VPP patch"]
+            p2mp["Lookup: {P2MP, src=outer_dst, dst=0.0.0.0}"]
```
yue-fred-gao, line 633: I think we should not reuse existing P2MP entry and adds P2MP lookup. This breaks existing P2MP + TEIB design, which wants to drop the packet if it doesn't match TEIB entries. Instead, I think we should introduce a new mode, for example, MP2P. It can be created the same way: create ipip tunnel src 10.0.0.1 mode mp2p. This creates a new entry for ipip tunnel lookup. You can then insert the lookup action after p2p or 6rd. If it is existing p2mp+teib, it won't hit the mp2p lookup because the entry is not there. I think this can be upstream to fd.io because it is a useful feature.

docs/HLD/vpp-ipiptunnel.md
lolyu: Yes, will add.

docs/HLD/vpp-ipiptunnel.md
lolyu: Good point, will add the APPL_DB section

docs/HLD/vpp-ipiptunnel.md
lolyu: encap and decap tunnel will be added separately.

docs/HLD/vpp-ipiptunnel.md
lolyu: Great idea, we sure can add a new tunnel type for our purpose. But naming it `MP2P` is pretty misleading here as `P2MP` means local is single-point and remote is multi-point, `MP2P` should means local is multi-point and remote is single-point.

How about name it `P2ANY`?

docs/HLD/vpp-ipiptunnel.md
lolyu: Good catch, it doesn't, will update.

docs/HLD/vpp-ipiptunnel.md
yue-fred-gao: I asked gemini.
```
The difference between MP2P (Multipoint-to-Point) and P2MP (Point-to-Multipoint) comes down to the direction of the traffic flow and how the network handles the duplication or aggregation of packets.
 
In short:
 
MP2P is Many-to-One (N:1). It is used for traffic aggregation.
P2MP is One-to-Many (1:N). It is used for traffic distribution or replication.
 
Here is a detailed comparison of the two concepts in the context of networking and tunneling:
 
 
1. MP2P (Multipoint-to-Point)
 
Traffic Flow: Many senders $\rightarrow$ One receiver.
How it works: Multiple remote endpoints send tunneled traffic to a single, centralized destination. The receiving node uses a single interface to accept and decapsulate traffic from all of these different sources.
Key Characteristic: Aggregation. The network paths merge as they get closer to the destination. The receiver does not need to maintain strict state for every sender; it just processes the incoming packets.
Common Use Cases:
Standard Unicast Routing: In MPLS or Segment Routing, traffic from all over the network destined for a specific router naturally merges into an MP2P flow.
Hub-and-Spoke VPNs: Spoke routers sending their internet or corporate traffic to a central Hub router (e.g., DMVPN Spokes sending to the Hub).
 
2. P2MP (Point-to-Multipoint)
 
Traffic Flow: One sender $\rightarrow$ Many receivers.
How it works: A single source sends a packet, and the network is responsible for delivering that exact same packet to multiple destinations. To do this efficiently, the network performs packet replication—copying the packet only when the path splits, rather than the sender transmitting individual copies to every receiver.
Key Characteristic: Replication (Multicast). The sender or the network fabric must maintain state about who the receivers are so it knows where to replicate and forward the packets.
Common Use Cases:
Multicast Traffic: Live video streaming, IPTV, or financial stock ticker feeds where one server broadcasts to thousands of viewers.
Routing Protocol Updates: A Hub router sending OSPF or EIGRP hello packets/updates to all of its Spoke routers simultaneously.
MPLS Traffic Engineering (RSVP-TE P2MP): Used to build a strict, engineered path from a single headend router to multiple tail-end routers, specifically to carry Multicast VPN (mVPN) traffic.
```
I think our use case fits MP2P. It describes sender->receiver relationship, not local->remote.

docs/HLD/vpp-ipiptunnel.md
yue-fred-gao: If sonic adds encap decap tunnel separately with same set of endpoints (to create a bidirectional tunnel), you will end up with calling vpp_ipip_tunnel_add twice with the same arguments, right? This is essentially how to solve the difference between sonic and vpp where the former creates 2 unidirectional tunnels but vpp always creates one bidirectional one.

docs/HLD/vpp-ipiptunnel.md
lolyu: Hi @yue-fred-gao, for decap, the receiver is local, the sender is remote; as p2mp means the sender is `mp` while the receiver is `p`, mp2p means the sender is `p` and the receiver is `mp`. This is not our situation, right?

yue-fred-gao: Hi @lolyu , from explanation from gemini, "2" is the direction of traffic ("2" is to). The role before "2" is sender and after is receiver. so p2mp means one sender to multiple receiver. mp2p means multiple sender to one receiver. Having said that, I am not particularly clinging to any term.

docs/HLD/vpp-ipiptunnel.md
lolyu: yes, this is a implementation detail and will have special handle.

docs/HLD/vpp-ipiptunnel.md
lolyu: yes, we need to refresh the interface list to get intf name to used by `interface_set_state`

lolyu:
Hi @yue-fred-gao, could you please help review again?

Few updates:
1. APPL_DB schema is added
2. dedup process is added to solve the VPP tunnel duplication due to SONiC programs encap/decap tunnels with the same src/dst IP.
3. add section to describe the solution to fix the tunnel unnumber failure due to rif not ready.

As for this PR: https://github.com/sonic-net/sonic-platform-vpp/pull/221, can we do this in two step, first let's unblock the SONiC decap with this hacky solution, and add a new ipip tunnel flag to support this afterwards?

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,854 @@
...
+    ecn_mode            = "copy_from_outer"/"standard"
+    ttl_mode            = "uniform"/"pipe"
+    encap_ecn_mode      = "standard"
+
+### TUNNEL_DECAP_TERM_TABLE
+    ; Stores a list of decap terms.
+    key                 = TUNNEL_DECAP_TERM_TABLE:tunnel_name:dst_ip    ; tunnel name:dst IP prefix as key
+    term_type   	      = "P2P"/"P2MP"                                  ; tunnel decap term type
+    src_ip              = source IP prefix                              ; present if term_type is "P2P"
+    subnet_type         = "vlan"/"vip"                                  ; the subnet type of the dst IP prefix, present
```
yue-fred-gao, line 191: what do the 2 subnet_type mean? How are they mapped 2 SAI attributes?

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,854 @@
...
+
+The **tunnel object** defines the P2P endpoints and QoS modes. For encap, the tunnel object carries `ENCAP_SRC_IP` (our loopback) which becomes the outer header source IP. The **nexthop's** `SAI_NEXT_HOP_ATTR_IP` provides the peer endpoint IP that becomes the outer header destination — this is the same value as the tunnel's `ENCAP_DST_IP` in the MuxOrch DualToR use case, but the VPP implementation reads it from the nexthop object.
+
+The **tunnel encap nexthop** is the actionable object — routes pointing to this nexthop will have their packets encapsulated through the tunnel. `MuxOrch` creates these nexthops for active-standby traffic steering.
+
+---
+
+## 5. SONiC Configuration
+
+### 5.1 Decap - APPL_DB State
```
yue-fred-gao, line 174: Thanks. Can you please add APPL_DB for encap as well?

yue-fred-gao:
> Hi @yue-fred-gao, could you please help review again?
> 
> Few updates:
> 
> 1. APPL_DB schema is added
> 2. dedup process is added to solve the VPP tunnel duplication due to SONiC programs encap/decap tunnels with the same src/dst IP.
> 3. add section to describe the solution to fix the tunnel unnumber failure due to rif not ready.
> 
> As for this PR: #221, can we do this in two step, first let's unblock the SONiC decap with this hacky solution, and add a new ipip tunnel flag to support this afterwards?

Thanks for the update. I think new ipip tunnel flag is not very complicated so I prefer you use the same commit to complete it. The reason is that for each new update to vpp (new patch or new version), we need to upload the prebuilt debs to buildkite so they are not rebuilt again. We want to minimize the versions in buildkite to limit the space usage.

lolyu:
> > Hi @yue-fred-gao, could you please help review again?
> > Few updates:
> > 
> > 1. APPL_DB schema is added
> > 2. dedup process is added to solve the VPP tunnel duplication due to SONiC programs encap/decap tunnels with the same src/dst IP.
> > 3. add section to describe the solution to fix the tunnel unnumber failure due to rif not ready.
> > 
> > As for this PR: #221, can we do this in two step, first let's unblock the SONiC decap with this hacky solution, and add a new ipip tunnel flag to support this afterwards?
> 
> Thanks for the update. I think new ipip tunnel flag is not very complicated so I prefer you use the same commit to complete it. The reason is that for each new update to vpp (new patch or new version), we need to upload the prebuilt debs to buildkite so they are not rebuilt again. We want to minimize the versions in buildkite to limit the space usage.

Hi @yue-fred-gao, thanks, one question: we only patches the downstream version, right? why do we need to upload debs to buildkite?

yue-fred-gao:
> > > Hi @yue-fred-gao, could you please help review again?
> > > Few updates:
> > > 
> > > 1. APPL_DB schema is added
> > > 2. dedup process is added to solve the VPP tunnel duplication due to SONiC programs encap/decap tunnels with the same src/dst IP.
> > > 3. add section to describe the solution to fix the tunnel unnumber failure due to rif not ready.
> > > 
> > > As for this PR: #221, can we do this in two step, first let's unblock the SONiC decap with this hacky solution, and add a new ipip tunnel flag to support this afterwards?
> > 
> > 
> > Thanks for the update. I think new ipip tunnel flag is not very complicated so I prefer you use the same commit to complete it. The reason is that for each new update to vpp (new patch or new version), we need to upload the prebuilt debs to buildkite so they are not rebuilt again. We want to minimize the versions in buildkite to limit the space usage.
> 
> Hi @yue-fred-gao, thanks, one question: we only patches the downstream version, right? why do we need to upload debs to buildkite?
@lolyu , we upload debs to buildkite to avoid building vpp from source every time. It saves CPU resources and make the sonic-vpp build more stable.

docs/HLD/vpp-ipiptunnel.md
```diff
@@ -0,0 +1,854 @@
...
+  if --refcount > 0:
+      return SUCCESS  // tunnel still in use
+  else:
+      erase from refcount map
+      interface DOWN + vpp_ipip_tunnel_del()
+```
+
+### 7.6 Deferred Unnumbered Interface Setup
+
+**Problem:** When an IPIP tunnel is created, it needs to borrow its IP address from the interface owning the tunnel source IP (e.g., `Loopback0`). However, the tunnel term may be created before the loopback RIF's IP address is programmed in VPP — `vpp_sw_interface_find_by_ip()` returns `-ENOENT`.
```
yue-fred-gao, line 572: how is ip address or unnumbered assigned to ipinip tunnel interface used in the forwarding? Is this for encap where ipinip tunnel interface as nexthop?

docs/HLD/vpp-ipiptunnel.md
lolyu: This maps to SAI MP2MP IPINIP tunnel, which is not covered here.

docs/HLD/vpp-ipiptunnel.md
lolyu: Hi @yue-fred-gao, the IPinIP encap tunnel is directly programed by orchagent as long as it is dualtor device, there is no APPL_DB config for this.

docs/HLD/vpp-ipiptunnel.md
lolyu: Yes, IP must be enabled on the ipip tunnel to allow the tunnel interface as an IP egress interface. The unnumbered assignment to the ipip tunnel makes the adjacency on the tunnel sw_if_index valid and ip-output works.

yue-fred-gao: I see. In vxlan encap, I assigned an IP from 0.0.0.0/8 to the vxlan tunnel intf for the same purpose.

docs/HLD/vpp-ipiptunnel.md
yue-fred-gao: got it

docs/HLD/vpp-ipiptunnel.md
yue-fred-gao: I see

yue-fred-gao (approved):

PR #225, opened 2026-04-27
sonic-platform-vpp trixie migration.

dypet:
Migrating platform vpp to Trixie. Corresponding build-image change: https://github.com/sonic-net/sonic-buildimage/pull/27013


rules/vpp.mk
```diff
@@ -1,6 +1,6 @@
 # libvpp package
 
-VPP_VERSION_BASE = 2510
+VPP_VERSION_BASE = 2513
```
yue-fred-gao: major version should follow vpp version in the upstream corresponding to the version we picked. Is it 2602 now?

dypet: Updated

yue-fred-gao (approved):
lgtm

yue-fred-gao (approved):

PR #228, opened 2026-05-05
Create vlan-bvi HLD

yue-fred-gao:
### why
VLAN BVI is not supported today and an issue has been created: https://github.com/sonic-net/sonic-buildimage/issues/26936

### what this PR does
Create an HLD to document how we plan to support ARP, DHCP and unicast packet to VLAN.


docs/vlan-bvi-hld.md
```diff
@@ -0,0 +1,355 @@
...
+
+### 5.2 Ingress: Untagged Member (e.g., Ethernet4 in Vlan10, untagged)
+
+```
+Wire (no tag) → bobm0 (dpdk-input, promisc on)
+  → ethernet-input: no tag → parent sw_if_index
+    → l2-input (BD 10)
+      → l2-input-vtr: no-op (no rewrite configured)
+        → l2-learn: learn src MAC on bobm0
+          → [LLDP/LACP?] → lcp-punt-l2-ethertype: punt to host (no flood)
```
lolyu: from section 3.3, lcp-punt-l2-ethertype is placed before l2-learn, here it shows after l2-learn?

docs/vlan-bvi-hld.md
```diff
@@ -0,0 +1,355 @@
...
+vppctl show lcp
+  bvi10     → tap_Vlan10
+  bobm0     → Ethernet0      (physical)
+  bobm0.10  → Ethernet0.10   (auto-subint)
+
+# Promiscuous mode
+vppctl show interface bobm0
+  flags: ... promisc ...
+
+# Punt L2 ethertypes (global)
```
lolyu: why arp punt-copy is missing here?

docs/vlan-bvi-hld.md
```diff
@@ -0,0 +1,355 @@
...
+a VPP sub-interface (e.g., `bobm0.10`), linux-cp automatically creates the
+corresponding Linux VLAN device (`Ethernet0.10`) on the host side. This
+eliminates the need for explicit `configure_lcp_interface()` calls for
+sub-interfaces.
+
+---
+
+## 4. Implementation Changes
+
+### 4.1 BVI LCP Pair + TC Filter
```
lolyu: In SONiC, vlanmgrd creates a Linux vlan netdev (e.g Vlan1000) and assigns it the gateway IP (e.g. 192.168.0.1/24). This is the sole authoritative ARP responder for the gateway address. ARP requests punted from VLAN member ports reach the kernel Vlan1000 interface, and the kernel replies normally. That's the intended control plane path.
This BVI LCP pair looks redundant in this context, all we need to do is the punting the arp packets to kernel on all vlan member interfaces, and let the kernel do its job.

lolyu (commented):
do we have plan to cover IPv6 NDP, DHCP?

docs/vlan-bvi-hld.md
yue-fred-gao: arp is not punted directly from member but through bvi

yue-fred-gao:
> do we have plan to cover IPv6 NDP, DHCP?

I changed the solution completely to punt dhcp directly via member interface with classifier, so for ARP and LLDP. IPv6 is slightly different. vpp has built-in support for ND with ip6-neighbor-discovery, which managed ND without involving control plane. But most likely is using the same classifier solution to punt ND to sonic, similar to ARP.

docs/vlan-bvi-hld.md
yue-fred-gao: outdated

docs/vlan-bvi-hld.md
yue-fred-gao: With the new solution, ARP is punted from member lcp interface

docs/HLD/vlan-bvi-hld.md
```diff
@@ -0,0 +1,576 @@
+# VLAN Bridge Domain with BVI — High-Level Design
+
+## 1. Problem Statement
```
lolyu, line 3: can you add examples on how does ASIC DB tables model those behaviors?

docs/HLD/vlan-bvi-hld.md
```diff
@@ -0,0 +1,576 @@
...
+      → l2-input (BD 10)
+        → l2-input-classify: etype=0x0806 → "other" table
+          → untag_other: etype=0x0806 (ARP) → HIT, opaque=1
+            → clone-on-hit:
+              CLONE → linux-cp-punt → Ethernet4 tap (untagged)
+                → kernel receives ARP on Ethernet4
+              ORIGINAL → continues L2 feature chain:
+                → l2-learn → l2-fwd → l2-flood → bvi10
+                  → bvi-input → arp-input → linux-cp (Vlan10)
+```
```
lolyu, line 613: Looks the kernel sees the same ARP reques twice, once on Ethernet4, once on Vlan100, and kernel will respond twice in this case.

For arp packets, they should not continue the L2 feature chain in this case. The wire between BVI and kernel Vlan interface should only forward L3 traffic, as stated in the problem statement.

yue-fred-gao: Yes, the remote host will see 2 arp replies but it doesn't do any harm though. ARP request needs to be flooded in the vlan bridge to maintain L3 connectivity between VLAN hosts. BVI is a member of the vlan bridge in vpp and it is a bigger change to remove it. So I decided to caveat this behavior. Do you see potential issues?

lolyu: There is **no SAI object that binds the ASIC VLAN RIF to the kernel VlanXXX netdev**, they are two independent objects that just happen to share the same MAC.

For the packet flow described in "### 5.7 IP Unicast to SVI (e.g., ping to Vlan10 IP)", the packet should be punted to the vlan member port (IP2ME), and reach the kernel Vlan10 netdev as they are in the same bridge.
This is the packet path on SONiC regardless of the platform, we should enforce the same behavior.

yue-fred-gao: They are independent objects but connected through TC.

It will be difficult to punt the unicast packet from member interface. In l2-input-classify node, we don't know if the packet is going to bvi or not, which is determined at ip4-lookup.
If we punt to the member in linux-cp-punt-xc node, vlan tag is already stripped. The change will be very invasive. On the other hand, what issue do you see with punting to Vlan10, other than it is not SONiC expected path?

lolyu: If the packet is punting to Vlan10, iptables will not work as expected, right?
Also it looks like Loopback devices are having the same issue - unicast packets (like icmp) are punted directly to the kernel Loopback netdevs. This is also an issue as dualtor sonic sends ICMP probes per Vlan member port and expects the ICMP replies from the Vlan member port.

yue-fred-gao: typically iptables work on ip interface (Vlan10 in this case), not member interface. By default, bridged frames do not pass through iptables. Do you have concrete example which feature applies iptables on bridge frames?
It is true that unicast to loopback interface is punted directly to the Loopback netdevs. This is the behavior in t1 and so far we don't see any issue in sonic-mgmt. If it is indeed an issue, it won't be solved by this HLD.
For the last one, ICMP will be punted through BVI tap and reply will be sent out from member interface. The remote will receive it on BVI. Does dualtor check which interface it receives it? Please share the code or design.

yue-fred-gao: I think I found the dualtor icmp probe: https://github.com/sonic-net/sonic-linkmgrd/blob/65f563308c689e3225fdf3fc249a132350e9879b/src/link_prober/LinkProberBase.cpp#L71.
The change will be quite invasive and might be difficult to get approval from fd.io.

yue-fred-gao:
Moved to https://github.com/sonic-net/SONiC/pull/2387.

PR #234, opened 2026-05-18
Add VPP patch 0010: opt-in inner-aware flow hash for IPinIP/GRE/NVGRE

yejianquan:
# Why
Support vpp inner packet hash compute aware for https://github.com/sonic-net/sonic-buildimage/issues/25762

This patch makes VPP's IPv4/IPv6 ECMP and bond LAG load-balancers hash on the inner header of IPinIP / 6in4 / 4in6 / 6in6 / GRE / NVGRE transit tunnels, so distinct inner flows between the same two tunnel endpoints distribute across paths and bond members instead of collapsing onto asingle path.

 ## Design

 Behaviour is **fully opt-in on both surfaces**:

 * **IP layer (ECMP)** — a new flow-hash bit `IP_FLOW_HASH_PEEK_INNER` (`1<<9`, CLI keyword `peek_inner`) in `flow_hash_config_t`. `IP_FLOW_HASH_DEFAULT` stays at `0x9F`; when the bit is not set, `ip4_compute_flow_hash` and `ip6_compute_flow_hash` behave exactly as upstream.

 * **LAG** — a **new** load-balance algorithm `BOND_API_LB_ALGO_L34_INNER = 6` in `src/vnet/bonding/bond.api` (CLI keyword `l34-inner`), backed by a **new** registered Ethernet hash function `hash-eth-l34-inner` (priority 50) in `src/vnet/hash/hash_eth.c`. The existing`BOND_API_LB_ALGO_L34 = 1` and `hash-eth-l34` are **byte-for-byte unchanged**, so any LAG that does not explicitly select `l34-inner` sees zero behaviour change and zero extra cost. This replaces the earlier in-place "always-on with safe fallback" edit of `hash-eth-l34` in response toupstream review (Fred Wang, lolyu).

 * **ABI compatibility** for the new enum value is preserved by the `[backwards_compatible]` annotation: `vppapigen` excludes the new member from the CRCs of `bond_create` / `bond_create2` / `sw_interface_bond_details` / `sw_bond_interface_details`, so a libsaivs built without the newvalue continues to link and run against a VPP built with it. The same mechanism is used by `enum sr_behavior` in `src/vnet/srv6/sr_types.api` and `enum ipsec_crypto_alg` in `src/vnet/ipsec/ipsec_types.api`.

 * **Shared helper** — `src/vnet/ip/ip_inner_aware_hash.h` houses the inner-peek inlines used identically by all three hash surfaces (`ip4_compute_flow_hash`, `ip6_compute_flow_hash`, `hash_eth_l34_inner`). Every entry point takes a `u32 remaining` bound; outer-fragment packets areskipped; IPv6 inner extension headers (HBH/DST/ROUTE) are walked, and Fragment / AH / ESP cause a fall-back to outer-only hashing.

 * **Out of scope** — VxLAN (RFC 7348 §4.2) and Geneve (RFC 8926 §3.3) place inner-flow entropy in the outer UDP source port; SRv6 (RFC 6437 / RFC 8754 §7) places it in the IPv6 flow label. The existing outer-only hash already distributes those classes when the ingress generates theprescribed outer entropy.

 ## Test coverage

 * **Unit tests** — `test/test_inner_aware_hash.py` covers IPv4/IPv6 outer × IPv4/IPv6 inner, NVGRE, fragmented outer, truncated outer, IPv6 inner extension headers, a determinism check, and a regression-safety class `TestLagL34LegacyOuterOnly` that pins down that `BOND_API_LB_ALGO_L34` / `hash-eth-l34` are byte-for-byte unchanged after this patch.

 * **Perf harness** — `test/test_inner_aware_perf.py` provides three packet-generator + `show runtime` measurement classes: `PEEK_INNER` on/off on ECMP, the new `L34_INNER` LAG algorithm with peek on, and the legacy `L34` LAG algorithm as a side-by-side A/B baseline.

 ## What this PR changes in this repo

 * `vppbld/patches/0010-sonic-inner-aware-flow-hash.patch` (new, 2849 lines, 13 files inside VPP, **+2326 / -52**)
 * `vppbld/patches/series` — adds the new patch with a 5-line scope comment listing in-scope encapsulations (IPinIP / 6in4 / 4in6 / 6in6 / GRE / NVGRE), out-of-scope ones (VxLAN / Geneve / SRv6), and the byte-for-byte-unchanged guarantee.

 ## How to verify

 Full `sonic-mgmt` `tests/fib/test_fib.py` suite on `vms-kvm-vpp-t1-lag` (regression **r13**):

 | Test | Pre-patch | Post-patch |
 |---|---|---|
 | `test_basic_fib`, `test_hash[ipv4]`/`[ipv6]`, `test_ecmp_group_member_flap` | PASS | PASS |
 | `test_ipinip_hash[ipv4]` / `[ipv6]` | FAIL (single path) | **PASS** |
 | `test_nvgre_hash[ipv4-*]` / `[ipv6-*]` | FAIL (single path) | **PASS** |
 | `test_vxlan_hash[ipv4-*]` / `[ipv6-*]` | FAIL (LAG collapse) | **PASS** |
 | `test_ipinip_hash_negative[ipv4]` / `[ipv6]` | PASS | PASS |

 Result: **16 passed, 698 warnings, 2 errors in 7264.48s (2:01:04)** — the 2 errors are syncd-shutdown teardown analyzer matches unrelated to flow hashing (they occur with or without this patch).

 ## Companion changes (separate PRs)

 * **HLD** — `docs/HLD/vpp-inner-aware-flow-hash.md` on branch `inner-aware-flow-hash-hld`
 * **libsaivs** — sonic-net/sonic-sairedis#1896: `vpp_create_lag()` selects `lb = VPP_BOND_API_LB_ALGO_L34_INNER` (`6`); `vpp_add_ip_vrf()` ORs `VPP_IP_API_FLOW_HASH_PEEK_INNER` into the default-VRF hash mask and adds the `AF_INET6 vpp_ip_flow_hash_set` call
 * **sonic-mgmt** — sonic-net/sonic-mgmt#24721: removes `test_ipinip_hash` / `test_nvgre_hash` / `test_vxlan_hash` from the `vpp` platform skip-list; adds a `vpp` `asic_type` branch for `test_nvgre_hash` `hash_keys`

vpp upstream PR: https://gerrit.fd.io/r/c/vpp/+/45864


yejianquan:
Will create PR to VPP upstream later

yejianquan:
vpp upstream PR: https://gerrit.fd.io/r/c/vpp/+/45864

lolyu (approved):
### 📊 Overview
**Files Changed**: 2 (`vppbld/patches/0010-sonic-inner-aware-flow-hash.patch`, rename from 0011→0010; `vppbld/patches/README.md` implied)
**Lines**: ~1927 in the patch file

VPP core patch implementing the inner-aware flow hash feature: new `IP_FLOW_HASH_PEEK_INNER` flag, shared `ip_inner_aware_hash.h` helper, changes to `ip4/ip6_compute_flow_hash`, `hash_eth_l34_inline`, VPP unit tests, and RST developer documentation.

### ✅ Strengths

- **`ip_inner_resolve` bounds checking** is thorough throughout — every byte access is gated against the `remaining` counter, and all failure paths (truncated packet, unknown inner IP version, fragmented inner, unsupported GRE protocol type, unrecognised IPv6 extension header) set `inner.valid = 0` and let the caller fall back to the outer-only hash. No crash path.
- **Fragment safety**: IPv4 outer fragments correctly skip peeking (`ip4_is_fragment` guard); IPv6 outer fragments set `outer_fragmented = 1` which blocks the `peek_inner` else-branch via the `!outer_fragmented` guard.
- **IP_FLOW_HASH_DEFAULT unchanged (0x9F)** — the new bit is at position 9, cleanly outside all existing defaults. Zero behaviour change for any FIB that does not explicitly request `peek_inner`.
- **ABI preserved**: no new `BOND_API_LB_ALGO_*` enumerator is introduced; `hash-eth-l34` is extended in place, so no enum regeneration or CRC bump is needed on the libsaivs side.
- **30+ VPP unit tests** covering the full tunnel matrix (IPv4/IPv6 outer × IPv4/IPv6 inner, NVGRE, fragmented outer, truncated outer, IPv6 extension-header walk, determinism) plus a perf harness — excellent coverage.

### 📝 Review Findings

#### 📝 Minor Issues

**`!PREDICT_FALSE(ip4_is_fragment(ip4))` idiom in `hash_eth_l34_inline`**

```c
if (!PREDICT_FALSE (ip4_is_fragment (ip4)))
{
    ...
    ip_inner_resolve (...);
}
```

The negation of `PREDICT_FALSE` is semantically correct but idiomatically unusual in VPP. The conventional expression is `PREDICT_TRUE(!ip4_is_fragment(ip4))` or simply `if (!ip4_is_fragment(ip4))` — both carry the same prediction hint that the non-fragment path is hot. Not a functional issue; purely a style nit.

**IPv6 outer in `hash_eth_l34_inline` does not check for outer fragmentation before calling `ip_inner_resolve`**

For the IPv6 LAG path, `ip_inner_resolve` is called directly with `ip6->protocol` without checking whether the outer IPv6 packet is a fragment (NH=44). In practice this is safe — if the outer IPv6 has a fragment extension header as its first next-header, `ip6->protocol` is 44, and `ip_inner_resolve` returns `valid=0` for that unknown protocol, falling back to the outer hash. However the IPv4 path *does* explicitly guard with `ip4_is_fragment`, so the asymmetry is slightly surprising. A comment noting why the IPv6 path omits the explicit guard (i.e., `ip_inner_resolve` handles NH=44 safely via the fallback) would help future readers.

#### 💡 Suggestions

- The RST developer doc (`inner_aware_hash.rst`) and the patch commit message mention SRv6 is not handled, but only implicitly (SRH/NH=43 is not listed in the "walk outer IPv6 extension headers" section). A one-line note in §Testing or the RST doc that SRv6 outer falls back to outer-only hash would complete the exclusion story documented in the accompanying HLD (PR #235).

### 🧪 Testing

30+ unit test cases covering the full functional matrix, plus a perf harness — this is more coverage than most VPP patches carry. No gaps found.

### Status
✅ Approved — clean implementation, thorough safety guards, minor style nit and one documentation suggestion only.

yue-fred-gao:
hi @yejianquan , have you run sonic-mgmt sanity with this change? Please share the test result to make sure no breakage.

vppbld/patches/0010-sonic-inner-aware-flow-hash.patch
```diff
@@ -0,0 +1,2849 @@
...
++	    {
++	      const ip6_fragment_ext_header_t *frag = (const ip6_fragment_ext_header_t *) cur;
++	      protocol = frag->protocol;
++	      cur += sizeof (ip6_fragment_ext_header_t);
++	      walked += sizeof (ip6_fragment_ext_header_t);
++	    }
++	  outer_fragmented = 1;
+ 	}
+-      else if (protocol == IP_PROTOCOL_TCP || protocol == IP_PROTOCOL_UDP)
++      if (protocol == IP_PROTOCOL_TCP || protocol == IP_PROTOCOL_UDP)
```
yue-fred-gao: Is this correct? If the packet is fragmented, we shouldn't hash L4 fields otherwise first packet and the rest of the fragments may go through different path due to different hash value.

vppbld/patches/0011-sonic-inner-aware-flow-hash.patch
```diff
@@ -0,0 +1,2849 @@
...
+ 	  is_tcp_udp = 1;
+-	  tcp = cur;
++	  is_udp = (protocol == IP_PROTOCOL_UDP);
++	  tcp = (const tcp_header_t *) cur;
++	}
++      else if (PREDICT_FALSE (peek_inner && !outer_fragmented && payload_length >= walked))
++	{
++	  u32 remaining = payload_length - walked;
++	  ip_inner_resolve (protocol, cur, remaining, &inner);
++	  if (PREDICT_FALSE (inner.valid))
```
yue-fred-gao, line 1031: can it be recursive to ip6_compute_flow_hash?

vppbld/patches/0010-sonic-inner-aware-flow-hash.patch
yejianquan: Thanks for pointing out! Enhanced the logic, please check again.

vppbld/patches/0011-sonic-inner-aware-flow-hash.patch
yejianquan: Thanks Fred -- this is worth thinking through.  We considered a recursive
 `ip6_compute_flow_hash` call and landed on the helper-based design
 (`ip_inner_resolve` + friends) for three concrete reasons:

 1. **The bond / LAG path has no `flow_hash_config_t`.**
    `hash_eth_l34_inner_inline()` (`src/vnet/hash/hash_eth.c:281`)
    takes `void **p` only -- VPP has no per-LAG hash-config concept
    at the ethernet layer.  A recursive `ip6_compute_flow_hash`
    call from there would have to synthesize a config out of thin
    air, and any change to IP-layer config semantics (SYMMETRIC,
    REVERSE, GTPV1_TEID) would silently bleed into LAG behavior.
    The helpers have no `flow_hash_config_t` dependency, so they
    plug into all three call sites (`ip4_compute_flow_hash`,
    `ip6_compute_flow_hash`, `hash_eth_l34_inner_inline`) unchanged.

 2. **Bounded budget vs untrusted inner `payload_length`.**
    `ip_inner_resolve` takes a `u32 remaining` derived from the
    *outer* header's `payload_length` clamped to the buffer size,
    and refuses to read past that limit regardless of what the
    inner header claims.  A recursive `ip6_compute_flow_hash`
    call would re-read the (untrusted) inner `payload_length`
    and use it as the new budget -- effectively an OOB-read
    primitive on truncated or adversarial packets.  Adding the
    same `remaining` parameter to `ip{4,6}_compute_flow_hash`
    itself would touch every existing caller.

 3. **Inner can be v4, GRE-IPv4, NVGRE-Eth-v4...**
    Recursing only covers the v6-in-v6 case; we'd still need a
    v4 / GRE / NVGRE dispatcher, so the helper-based
    `ip_inner_resolve` doesn't go away.

 The underlying DRY concern is real, though: the v6 extension-
 header walk does live in two places.  We factored the common
 part into `ip_inner_v6_walk_ext_headers()`, which is currently
 called from `ip_inner_resolve_v6` only -- in principle we could
 also call it from `ip6_compute_flow_hash` itself, but the two
 walkers have intentionally different policies (the outer walker
 keeps going past a FRAG header and sets `outer_fragmented` so
 the caller can still hash on src/dst; the inner walker refuses

vppbld/patches/0011-sonic-inner-aware-flow-hash.patch
```diff
@@ -0,0 +1,2911 @@
...
++      /* Fragmentation invariant: a packet that carries a Fragment ext
++       header IS itself one fragment of the original L4 datagram --
++       never a complete L4 message.  Only fragment-0 carries the real
++       TCP/UDP src/dst ports; later fragments carry arbitrary
++       inner-payload bytes at the same buffer offset.  Hashing those
++       bytes as ports would send fragments of the same flow down
++       different paths and break reassembly at the destination.
++       Thus we must NOT read tcp->src/dst when outer_fragmented is
++       set.  This restores upstream pre-patch behaviour.  */
++      else if (!outer_fragmented && (protocol == IP_PROTOCOL_TCP || protocol == IP_PROTOCOL_UDP))
```
yejianquan, line 1020: Hi @yue-fred-gao  here's the updated logic
added comments and new ut to cover this scenario.

yejianquan:
> hi @yejianquan , have you run sonic-mgmt sanity with this change? Please share the test result to make sure no breakage.
Hi @yue-fred-gao , yeah, all of them passed.

sonic-mgmt fib/test_fib.py results
 ====================================================
 PASS    test_basic_fib[True-True-1514]
 PASS    test_hash[ipv4]
 PASS    test_hash[ipv6]
 PASS    test_ipinip_hash[ipv4]
 PASS    test_ipinip_hash[ipv6]
 PASS    test_ipinip_hash_negative[ipv4]
 PASS    test_ipinip_hash_negative[ipv6]
 PASS    test_vxlan_hash[ipv4-ipv4]
 PASS    test_vxlan_hash[ipv6-ipv6]
 PASS    test_vxlan_hash[ipv6-ipv4]
 PASS    test_nvgre_hash[ipv4-ipv4]
 PASS    test_nvgre_hash[ipv4-ipv6]
 PASS    test_nvgre_hash[ipv6-ipv6]
 PASS    test_nvgre_hash[ipv6-ipv4]
 PASS    test_vxlan_hash[ipv4-ipv6]    (body PASS, teardown ERROR*)
 PASS    test_ecmp_group_member_flap[True-False-1514]    (body PASS, teardown ERROR*)
 ====================================================
 16 testcase bodies passed, 0 failed, 0 skipped.

 (*) Both teardown ERRORs are infra-only: loganalyzer's
 parallel_run fixture failed because core_uploader.py could
 not reach the corefile collector proxy (-
 timeout) to upload a pre-existing core file
 syncd.1778467508.68.core dated 2026-04-09 — well before
 this patch was deployed. The patch under test did not
 generate any new core files (no fresh entries in /var/core
 during the 2-hour run). These errors are unrelated to the
 patch and have been seen in prior unrelated runs on
 vlab-vpp-01.

Post-test sanity check: 5 passed, 1 skipped, 0 failed (no DUT regressions).

DUT verification artifacts (collected immediately after r13):

 - vppctl show bond: 8 BondEthernets all on l34-inner ✓
 - vppctl show hash: both hash-eth-l34-inner (new) and hash-eth-l34 (legacy, unchanged) registered ✓
 - vppctl show ip fib: peek_inner set on VRF:0 v4 ✓
 - BGP: 24/24 Established ✓
 - No new cores in /var/core/ ✓

vppbld/patches/0011-sonic-inner-aware-flow-hash.patch
yue-fred-gao: Good point

yue-fred-gao (approved):

yue-fred-gao:
Have you created fd.io PR? I think this is a valuable contribution. If it is merged, we don't have to deal with merge conflicts in the future. Please add the fd.io PR to the description.

yue-fred-gao:
please resolve the conflicts

yejianquan:
> Have you created fd.io PR? I think this is a valuable contribution. If it is merged, we don't have to deal with merge conflicts in the future. Please add the fd.io PR to the description.

Hi @yue-fred-gao , yeah: https://gerrit.fd.io/r/c/vpp/+/45864 all the changes are cherry-picked to the PR.
have resolved the conflict

PR #236, opened 2026-05-25
docs: add VPP route flow counter HLD

cyw233:
Add VPP route flow counter HLD

### DIFF
--- a/docs/HLD/vpp-route-flow-counter.md
+++ b/docs/HLD/vpp-route-flow-counter.md
@@ -1,0 +1,214 @@
+# VPP Route Flow Counter HLD
+
+## Table of Contents
+
+1. [Revisions](#revisions)
+2. [Scope](#scope)
+3. [Definitions and Abbreviations](#definitions-and-abbreviations)
+4. [Background](#background)
+5. [Problem Statement](#problem-statement)
+6. [Proposed Solution](#proposed-solution)
+7. [Design Details](#design-details)
+8. [Code References](#code-references)
+9. [Expected Behavior](#expected-behavior)
+10. [Risks and Open Questions](#risks-and-open-questions)
+11. [Review Focus](#review-focus)
+
+## Revisions
+
+| Rev | Date | Author(s) |
+|-----|------|-----------|
+| v0.1 | 05/25/2026 | Chenyang Wang |
+
+## Scope
+
+This document describes the high level design for route flow counter support on the SONiC VPP platform.
+
+The change is implemented in the VPP virtual switch implementation in `sonic-sairedis`, under `vslib/vpp`. It enables SONiC route flow counters to be backed by VPP route statistics so tests and operators can verify that counters increment when traffic hits matched routes.
+
+### In scope
+
+- Capture VPP route `stats_index` when routes are programmed.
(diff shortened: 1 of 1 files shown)


docs/HLD/vpp-route-flow-counter.md
```diff
@@ -0,0 +1,213 @@
...
+std::map<sai_object_id_t, std::map<sai_stat_id_t, uint64_t>> m_routeCounterStatsBaseMap;
+```
+
+When a counter is bound to a route, the current VPP route stats are captured as the base. Normal reads return the delta from that base. Clear and read-and-clear modes update the base to the current VPP value.
+
+The code also handles a possible VPP route stats reset by treating a large wrap/reset delta as zero and updating the base.
+
+### Route updates
+
+Route updates can reprogram the route in VPP and may produce a new stats index. The update path deletes the old VPP route, adds the new VPP route, refreshes the cached stats index, and refreshes the base for any bound route counter.
```
yue-fred-gao: This means the flow counter may go down, right? Are we going to caveat if this is not handled?

docs/HLD/vpp-route-flow-counter.md
```diff
@@ -0,0 +1,213 @@
...
+
+Rollback is attempted if reprogramming or internal state update fails.
+
+### Next-hop group member changes
+
+Routes using next-hop groups can be affected when a next-hop group member is added or removed. The VPP SAI layer refreshes the route stats index for impacted routes after the path update. If the route has a bound counter, the route counter base is reset to the current VPP stats for the new stats index.
```
yue-fred-gao: same as above

docs/HLD/vpp-route-flow-counter.md
cyw233: @yue-fred-gao Yes, with the current design the visible flow counter may go down after route reprogramming or next-hop group member changes, because VPP may return a new route stats index and we currently reset the route counter base to the current value of that new index.

Example: before an NHG change, the old VPP stats index has `current=1500, base=1000`, so the visible counter is 500. After the NHG change, VPP returns a new stats index with `current=20`, and we reset `base=20`, so the visible counter becomes 0. That means the counter can appear to drop from 500 to 0 even though the user did not explicitly clear it.

Do you think it is acceptable to caveat this behavior in the HLD, or should we preserve the logical counter value in the `sonic-sairedis` implementation? One possible fix is to carry forward the previously visible delta across stats index refreshes, e.g. keep `carried_delta=500`, then report `carried_delta + (new_current - new_base)`. In the example above, the counter would remain `500 + (20 - 20) = 500`, and future packets would continue increasing from there.

yue-fred-gao: I feel it is not a big change to keep carried_delta so it is better to implement it instead of caveat.

cyw233: @yue-fred-gao I updated the HLD and also updated (and verified) the implementation in https://github.com/sonic-net/sonic-sairedis/pull/1894. Please review again, thanks

yue-fred-gao (approved):
lgtm

PR #237, opened 2026-05-25
[linux-cp/dpdk] Fix sub-port and LAG sub-port datapath on sonic-platform-vpp

### LINKED ISSUE #227: PortChannel/LAG subinterfaces are not working on SONiC-VPP
PortChannel/LAG subinterfaces are not working on SONiC-VPP.

This was reproduced by enabling the sonic-mgmt `port_in_lag` parameter for the sub-port forwarding test:

```text
sub_port_interfaces/test_sub_port_interfaces.py::TestSubPorts::test_packet_routed_with_valid_vlan[port_in_lag]
```

Environment from that run:

```text
Topology: t1-lag-vpp
Platform: kvm
ASIC type: vpp
DUT: vlab-vpp-01
```

Observed result:

```text
test_packet_routed_with_valid_vlan[port_in_lag] failed on both the original attempt and retry.

AssertionError: Did not receive expected packet on any of ports [16] for device 0.
```

Expected behavior:

`PortChannelX.<vlan>` / LAG subinterfaces should be programmed in VPP and should route the valid VLAN packet, the same way the test expects for regular sub-port interfaces.

Actual behavior:

The valid VLAN packet is not routed out to the expected PTF port for `port_in_lag`.

Source-level suspicion:

The RIF SUB_PORT create path appears to accept LAG parents, but still calls `create_sub_interface(tap_to_hwif_name(dev), vlan_id, vlan_id)` where `dev` is `PortChannelX`. `tap_to_hwif_name()` is backed by `sonic_vpp_ifmap.ini`, which is generated with physical `Ethernet*` mappings only, so `PortChannelX` does not resolve to the VPP `BondEthernetX` hwif and the VPP subinterface is not created on the correct parent.

This issue is specifically for `PortChannelX.<vlan>` / LAG subinterfaces not being programmed/routed correctly in SONiC-VPP.

lunyue-ms:
### Description of PR

Summary:
Three platform-level fixes that, together with the companion
`sonic-sairedis` PR, restore the sub-port and LAG-sub-port datapath
on `sonic-platform-vpp`. After this PR, PTF
`test_sub_port_interfaces.py::TestSubPorts::test_packet_routed_with_valid_vlan`
passes on `vlab-vpp-01` for both `[port]` and `[port_in_lag]`
parametrisations.

Fixes: [sonic-net/sonic-platform-vpp#227](https://github.com/sonic-net/sonic-platform-vpp/issues/227)

Companion PR:
- https://github.com/sonic-net/sonic-sairedis/pull/1907 (must be merged together to reproduce
  the PASS result)

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation update
- [ ] Test improvement

On `sonic-platform-vpp`, three independent issues keep the sub-port
datapath broken: VPP linux-cp holds the sub-port host netdev DOWN after
each itf-pair (re)create, DPDK virtio ports come up non-promiscuous and
drop broadcast ARP on freshly-created L3 sub-ports, and
`lcp-auto-subint` races sairedis to install the LAG sub-port LCP pair
(producing the wrong host tap). Together with the companion sairedis
changes, fixing these three restores sub-port and LAG sub-port
datapath end-to-end.

#### How did you do it?

Changes are confined to `docker-syncd-vpp` / `docker-sonic-vpp` config
and `docs/HLD/`:

1. **`lcp-sync` enabled** — VPP propagates sub-interface admin state to
   the host tap; without it linux-cp keeps the host netdev DOWN after
   itf-pair (re)create, even if IntfMgr already ran `ip link set up`.
2. **DPDK promisc reconciler** — dedicated supervisord program
   `vpp_promisc.sh` polls for DPDK port appearance and runs
   `vppctl set interface promiscuous on bobm*`. Real ASICs are
   implicitly promiscuous on routed ports; VPP's `startup-config`
   directive cannot fix this because it runs before the DPDK plugin
   initialises ports.
3. **`lcp-auto-subint` disabled** — sairedis is now the sole producer
   of sub-port LCP pairs and atomically wires up
   `configure_lcp_interface(BondEthernet<id>.<vlan>, be<id>.<vlan>)`
   plus the bidirectional `tc mirred` bridge between `be<id>.<vlan>`
   and the SONiC L3 netdev `PortChannel<id>.<vlan>` in the same SAI
   RIF create.
4. **Promisc reconciler hardening** — phase-1 timeout exits 0 with
   `autorestart=unexpected` (no respawn storm on virtio-only platforms),
   and a whitespace-tolerant regex avoids re-issuing
   `set promiscuous on` every 2s on minor VPP output format changes.
5. **HLD docs** — `vpp-lag.md` gains a new "LAG Subinterface" section
   describing the 3-tuple naming (`PortChannel<id>.<vlan>` /
   `BondEthernet<id>.<vlan>` / `be<id>.<vlan>`) and the bidirectional
   `tc mirred` bridge; `SONICVPP-HLD.md` adds a LAG Subinterface row
   to the "Interface Create Flow by Type" table.

#### How did you verify/test it?

Result: **22 passed / 10 failed in ~32 min**.

Failure breakdown:

* **6 upstream-blocked**, out of this PR's scope:
  * SVI variants (×2): `test_routing_between_sub_ports_and_port[*-svi]`
    — blocked by VLAN BVI / SVI routing not implemented in `vslib/vpp`
    (sonic-buildimage #26936).
  * IPinIP variants (×4): `test_tunneling_between_sub_ports[*]`
    — blocked by sairedis PR #1860 (IPinIP encap/decap; VPP 2510 enum
    needs `MP2P->MP` alias).
* **4 LAG-sub-port follow-up**, also out of scope:
  `test_routing_between_sub_ports_and_port[port_in_lag-l3]`,
  `test_routing_between_sub_ports_unaffected_by_sub_ports_removal[port_in_lag-different]`,
  `test_balancing_sub_ports[port_in_lag]`,
  `test_max_numbers_of_sub_ports[port_in_lag]`.
  DUT-originated ARP on a LAG sub-port can still fail to install a
  neighbor into the VPP sub-interface RIF in these stress / cross-port
  cases. The `[port]` counterparts PASS in the same run, isolating the
  bug to the LAG-specific plumbing (tc bridge or kernel → VPP
  propagation on top of it). Tracked as a follow-up PR.

#### Any platform specific information?

Changes only touch `sonic-platform-vpp` (`docker-syncd-vpp` /
`docker-sonic-vpp` config + HLD). No impact on other ASIC platforms.
Both syncd and sonic images are updated symmetrically. The companion
`sonic-sairedis` PR is required to reproduce the PASS result.

### Documentation

Updated `docs/HLD/vpp-lag.md` (new "LAG Subinterface" section, v0.3
revision, status row) and `docs/HLD/SONICVPP-HLD.md` (Subinterface row
aligned with current behaviour; new LAG Subinterface row).


docker-sonic-vpp/conf/vpp_promisc.conf
yue-fred-gao, line 1: I am not sure about this. vpp has api to do that. in vlan-bvi HLD (https://github.com/sonic-net/sonic-platform-vpp/pull/228), it has this.
### 4.7 Promiscuous Mode on Every Physical Interface

**File**: `SwitchVppHostif.cpp`

At LCP pair creation for each physical port:
```cpp
configure_lcp_interface(hwif_name, dev, true);
interface_set_promiscuous(hwif_name, true);  // <-- added
```

docker-sonic-vpp/conf/startup.conf.tmpl
```diff
@@ -296,5 +296,12 @@ plugins {
 
 # Why not support sub interfaces
 linux-cp {
-   lcp-auto-subint
```
yue-fred-gao, line 304: If we disable it, vpp sai needs to create sub interface for regular Ethernet interface. Is this in the plan?

docker-sonic-vpp/conf/startup.conf.tmpl
```diff
@@ -296,5 +296,12 @@ plugins {
...
+   # lcp-auto-subint disabled: auto-pair "BondEthernet<id>.<vlan>" <-> "be<id>.<vlan>"
```
yue-fred-gao: This is a big change. Please make sure you run full sonic-mgmt tests to avoid breakage

docker-sonic-vpp/conf/startup.conf.tmpl
lunyue-ms: In https://github.com/sonic-net/sonic-sairedis/pull/1907,   In SwitchVppRif.cpp::vpp_create_router_interface():
     int ret = create_sub_interface(parent_hwif, vlan_id, vlan_id);
     if (ret != 0) { return SAI_STATUS_FAILURE; }
 
     ret = configure_lcp_interface(hw_subifname, lcp_host, true);
     if (ret != 0) {
         delete_sub_interface(parent_hwif, vlan_id);   // rollback
         return SAI_STATUS_FAILURE;
     }

For a regular Ethernet sub-port (e.g. Ethernet0.10) this will be:
     create_sub_interface("bobm0", 10, 10)
     configure_lcp_interface("bobm0.10", "Ethernet0.10", true)

So I think that it covers to create sub interface for regular Ethernet interface.

docker-sonic-vpp/conf/startup.conf.tmpl
lunyue-ms: Yes, it is. I wil run a roll of ful regression tests to make sure it doesn't break any other tests.

docker-sonic-vpp/conf/vpp_promisc.conf
lunyue-ms: Good idea. interface_set_promiscuous() is listed in HLD #228 but isn't in master yet — the underlying VPP API sw_interface_set_promisc exists, so it's a thin shim.
 
I'll add it into sairedis PR #1907: add the wrapper, call it from SwitchVppHostif::createHostif after configure_lcp_interface(), and remove vpp_promisc.sh from this PR. Do you agree?

lunyue-ms (commented):
Thanks @yue-fred-gao for your comments. Please check if my thoughts is feasible.

docker-sonic-vpp/conf/startup.conf.tmpl
lunyue-ms: I ran additional regression testing and found that `arp/test_neighbor_mac.py` is failing intermittently with this change set. 
Initial debugging shows the failure is not caused by VPP DPDK promisc being off — `bobm0` stays promiscuous during the failure window. Packet capture indicates that in failing cases the ARP reply for `20.0.0.1` is sent with the Linux/LCP  tap MAC instead of the expected port/router MAC, so the following ICMP ping from PTF fails. I am still debugging the LCP/hostif MAC handling and will update the PR once I have a fix or a clearer root cause.

yue-fred-gao:
I agree we disable lcp-auto-subint. I see some cases (with VLAN bridge) subinterface is not desirable

lunyue-ms:
@yue-fred-gao  I included the changes of platform_vpp and sairedis. The regression test got  passed. So, "disable lcp-auto-subint" + "enable lcp-sync" will not break other tests.
https://elastictest.org/scheduler/testplan/6a21cd1595ed954f0d3d700e?searchTestCase=&testcase=arp%2Ftest_neighbor_mac.py&type=log

Since it replys on https://github.com/sonic-net/sonic-sairedis/pull/1907 to make sure other tests get passed, can you help to review that PR before merging this one(#237)? Thanks!

docs/HLD/vpp-lag.md
```diff
@@ -132,6 +136,31 @@ tc filter add dev <bond-tap-interface> parent ffff: \
...
 <a id="item-8"></a>
+## LAG Subinterface
+
+PortChannel subinterfaces are tagged L3 interfaces on top of a LAG (`PortChannel<id>.<vlan>`), used for example by the t1-lag topology where every LAG carries multiple `/30` peering subnets, one per VLAN.
+
+A LAG subinterface has three coupled names that must stay consistent: `PortChannel<id>.<vlan>` (the SONiC L3 kernel netdev that teamd creates and that the IP lives on), `BondEthernet<id>.<vlan>` (the VPP sub-interface on the bond), and `be<id>.<vlan>` (the LCP host, a VLAN child of the dummy tap `be<id>`). On `SAI_ROUTER_INTERFACE_TYPE_SUB_PORT` create, `vpp_get_hwif_name` must return `BondEthernet<id>.<vlan>`; returning the parent bond name causes every subsequent SAI call on the sub-interface (IP add, neighbor add, delete) to operate on the wrong VPP object.
+
+### Plumbing
+
+For a non-LAG sub-port (`Ethernet0.10`) the LCP host can simply be the SONiC L3 netdev itself, because it sits on top of a VPP-created tap (`tap<N>`). That does not work for a LAG sub-port: `PortChannel<id>.<vlan>` is a teamd VLAN child, not a VPP tap, and VPP linux-cp can only punt to a kernel netdev backed by a VPP tap.
```
yue-fred-gao: This solution probably works but I don't think it is same as other hw platforms. Does sonic expects packets to be punted from member interfaces instead of directly to the port channel interface? Same for outgoing path. The way implemented here bypasses linux bond group.

docs/HLD/vpp-lag.md
```diff
@@ -132,6 +136,31 @@ tc filter add dev <bond-tap-interface> parent ffff: \
...
+
+The design pairs the VPP sub-interface with `be<id>.<vlan>` (a VLAN child of the dummy tap `be<id>` that already exists for the parent LAG), and then bridges that LCP host to the SONiC L3 netdev `PortChannel<id>.<vlan>` with a bidirectional `tc mirred` redirect. The ingress redirect (`be<id>.<vlan>` → `PortChannel<id>.<vlan>`) carries VPP-punted ARP/ICMP frames to the netdev that holds the IP, so the kernel handles them normally. The egress redirect (`PortChannel<id>.<vlan>` → `be<id>.<vlan>`) carries kernel-originated frames into VPP via the LCP tap, where the linux-cp cross-connect forwards them to `BondEthernet<id>.<vlan>`. Both directions are mandatory: a one-direction-only bridge silently breaks LAG sub-port ping, and failure of either direction rolls back the LCP pair and VPP sub-interface so the SAI RIF create fails cleanly.
```
yue-fred-gao: Does it not work without egress redirect? packets sent to the pc sub interface will be added with vlan tag then to the bond driver. bond driver sends it to a member interface, which is injected to vpp and transmitted from a front panel port through linux cp cross connect.

yue-fred-gao: no it doesn't. I think the problem is the member interface doesn't have sub interface. Packet injected to the member interface will be dropped with "unknown vlan". I am implementing a new node that bypasses ethernet-input to send the packet directly from egress port. I think this fits sonic design. Packets exits from EthernetX netdev are fully formatted with proper l2 header. They can be sent out directly from physical port without further processing. Is this correct?

lunyue-ms: Thanks a lot. Please correct me if my understanding is wrong:
I think that your idea is a proper one and follows the sonic design principle. So the new node can solve the TX direction via the following path. Is it?
PortChannelX.Y -> Linux bond/team -> Ethernet member -> VPP new direct-TX node -> wire

For the RX direction of host/CP packets, do you expect to keep the current design?
VPP BondEthernetX.Y -> beX.Y -> tc -> PortChannelX.Y
Or are you also planning a different RX path?

docs/HLD/vpp-lag.md
lunyue-ms: Thanks for the point. Yes, the current modification bypasses the linux bond group by redirecting PortChannelX.Y directly to the beX.Y.  This is a workaround. I think that your idea(adding a vpp node to handle TX packets) in the following comments is the right way.

docs/HLD/vpp-lag.md
lunyue-ms: Hi @yue-fred-gao , I have removed the  tc mirred redirection part in sairedis code and HLD. Now data tests related to [port_in_lag-xxx] don't pass because ARP cannot reach PortChannel<id>.<vlan>, so the neighbor never resolves. I will skip them in sonic-mgmt for now (plain [port-xxx] still passes); we can unskip once your new node lands.

yue-fred-gao: Thanks for the update. I am thinking if we should change portchannel punt path. It is working so far but probably doesn't match what sonic is intended and how other platforms work. The tx path above is correct. For rx, I think the new design is VPP BondEthernetX.Y -> VPP tap_BondEthernetX -> (vlan tag is restored) -> Ethernet member -> PortChannelX -> (vlan tag is stripped) -> PortChannelX.Y
With this design, we don't need to create VPP tap_BondEthernetX.Y. Even tap_BondEthernetX is created but never used for sending the traffic so tc is not needed. This is the BVI path in HLD https://github.com/sonic-net/SONiC/pull/2387. It should be a small change to extend it to bond.

lunyue-ms: Thanks @yue-fred-gao , your design is the ideal solution and is consistent with the BVI path. For now I think it's best for me to leave this ticket here. Once the new node is implemented, supporting the bond interface should require very little change on the sairedis side.

PR #241, opened 2026-05-27
Add vpp NAT HLD

yue-fred-gao:
### why
sonic-vpp doesn't support NAT feature while vpp does. This HLD outlines the NAT feature in SONiC and that in vpp. What are the mismatch between them and how to align them to support NAT feature in sonic.

### what this PR does
Create a HLD with detailed design and implementation schedule.


yue-fred-gao:
moved to SONiC/doc/vpp

PR #244, opened 2026-06-01
syncd-vpp: keep --privileged

xq9mend:
A companion PR to sonic-net/sonic-buildimage#27522 removes `--privileged` from `$(DOCKER_SYNCD_BASE)_RUN_OPT` in all syncd template mk files, replacing it with specific capabilities (`--cap-add=SYS_RAWIO --cap-add=SYS_ADMIN --cap-add=NET_ADMIN`).

docker-syncd-vpp inherits from the bookworm template, so it would lose `--privileged` as a side effect. Since docker-syncd-vpp is not used in production and VPP testing environments depend on full device access, this PR explicitly re-adds `--privileged` in the platform mk file to preserve the existing behavior.


xq9mend:
@yejianquan @lolyu @dypet

yue-fred-gao (approved):

PR #248, opened 2026-06-03
Build both Bookworm and Trixie

saiarcot895:
sairedis, swss, and swss-common repos still need Bookworm artifacts, because docker-sonic-vs is still on Bookworm. This PR _should_ allow building both Bookworm and Trixie artifacts.

Note that artifact names change with this PR, and so will require changes to downstream consumers.


saiarcot895:
@dypet @yue-fred-gao FYI.

sairedis, swss, and swss-common repos still need Bookworm artifacts, because docker-sonic-vs is still on Bookworm. This PR _should_ allow building both Bookworm and Trixie artifacts.

yue-fred-gao (approved):

PR #250, opened 2026-06-05
docs: add VPP LAG member egress HLD

cyw233:


docs/HLD/vpp-lag-member-egress-disable.md
```diff
@@ -0,0 +1,379 @@
...
+
+## 7. High Level Design
+
+### 7.1 Design Summary
+
+SONiC-VPP maps the SAI egress-disable state to VPP bond membership:
+
+| SAI egress-disable state | VPP action |
+|---|---|
+| `true` | Keep the LAG member object in SAI state, but remove the port from the VPP bond |
```
yue-fred-gao, line 106: Instead of remove the port from vpp bond, can we set the port to down? I feel it is a lot of churns to add/remove port from bond.

docs/HLD/vpp-lag-member-egress-disable.md
```diff
@@ -0,0 +1,379 @@
...
+| `false` | Add the port to the VPP bond if it is not already present |
+
+This keeps the object lifecycle and dataplane programming separate:
+
+- The SAI LAG member object can exist even when egress is disabled.
+- The VPP bond member exists only when the member is eligible for egress traffic.
+
```
lolyu, line 113: This is an honest and important disclosure — but for an HLD it should be *justified*, not just stated. `SAI_LAG_MEMBER_ATTR_EGRESS_DISABLE` is by SAI definition an **egress** attribute, yet this design also drops the member from **ingress** as a side effect of full bond detachment. Worth an explicit sentence on why that divergence from strict SAI semantics is acceptable (e.g. SONiC continues running LACP so the member isn't selected for ingress distribution anyway, or downstream behavior tolerates it). A strict-SAI reader will flag this exact point, so addressing it here closes the gap pre-emptively. This also ties directly into @yue-fred-gao's churn question: the ingress side effect is part of the cost of modeling egress-disable as full detach vs. admin-down.

docs/HLD/vpp-lag-member-egress-disable.md
```diff
@@ -0,0 +1,379 @@
...
+```text
+SONiC PortChannel
+    -> SAI LAG
+        -> VPP BondEthernet
+
+SONiC PortChannel member
+    -> SAI LAG member
+        -> VPP bond member
+```
```
lolyu, line 62: Minor wording: "a stale generated MAC address instead of the member or bond MAC address" — "the member or bond MAC" is ambiguous about which is the intended correct value. Suggest naming the specific one the tap should inherit (member MAC, I believe, for the first-member case) so the rationale for the delayed-tap design is unambiguous.

docs/HLD/vpp-lag-member-egress-disable.md
```diff
@@ -0,0 +1,379 @@
...
+If all enabled members are later disabled after the LCP/tap pair already exists, this feature does not remove the LCP/tap pair immediately. The LAG can temporarily have an existing LCP/tap pair and no enabled VPP bond members. Re-enabling a member reuses the existing LCP/tap pair, and normal LAG removal remains responsible for LAG-level cleanup.
+
+### 7.8 Remove Path and Lifecycle Edge Cases
+
+The LAG member remove path is not redesigned by this feature. The expected lifecycle behavior is:
+
+| Case | Expected behavior |
+|---|---|
+| Remove an enabled member | Remove it from the VPP bond, then remove the SAI object state |
+| Remove a disabled member | Remove the SAI object state; VPP bond membership is already absent and should be treated as a no-op or tolerated by the remove path |
```
lolyu, line 247: §7.8 calls out "Deleting a disabled member is an important regression scenario because the member may already be absent from the VPP bond" — good catch — but the Error Handling table (§10) has no corresponding row for it. Since the doc itself flags this as the important regression case, the error-handling contract should be explicit: does `vpp_remove_lag_member()` return success (no-op) when the member is already absent from the bond? Tightening the loop between §7.8 and §10 would make the lifecycle contract complete.

docs/HLD/vpp-lag-member-egress-disable.md
```diff
@@ -0,0 +1,379 @@
...
+  v
+SwitchVpp::setLagMember(EGRESS_DISABLE=true)
+  |
+  v
+getLagMemberEgressDisableAction() -> DETACH
+  |
+  v
+vpp_remove_lag_member()
+  |
+  v
```
lolyu, line 322: Suggest adding a row here for the "remove a member already absent from the VPP bond" case described in §7.8 (the disabled-member-delete regression scenario). Right now that path's expected status/handling isn't captured in this table even though §7.8 names it as the important edge case.

lolyu (commented):
**Review:** https://github.com/sonic-net/sonic-platform-vpp/pull/250

A well-structured, genuinely thorough HLD — clear call flows, a behavior matrix, and an error-handling table. I verified its fidelity to the implementation: the methods it describes (`setLagMember`, `getLagMemberEgressDisableAction`, `vpp_ensure_lag_lcp`, the no-op/detach/add decision) match sonic-sairedis #1928, and both relative doc links (`vpp-lag.md`, `SONICVPP-HLD.md`) resolve to existing files in `docs/HLD/`. Nice work.

Non-blocking (COMMENT). Two substantive suggestions, both about making the design *rationale* explicit rather than the design itself:

1. **Record the "detach-from-bond vs. set-member-down" tradeoff as a first-class decision.** @yue-fred-gao already raised this in-thread (churn of add/remove vs. admin-down). §7.1 actually contains the answer — "VPP does not expose a SONiC-VPP helper for an egress-only disable flag" — but it's buried in prose. An HLD is exactly where this belongs as an **Alternatives Considered** subsection: the admin-down option, why it wasn't used (VPP API limitation), and the churn cost. That pre-empts the same question from every future reader.

2. **Justify the ingress side effect, don't just state it** (see inline). The rest are inline polish nits. Overall this is solid and accurately documents the feature.

cyw233:
Moved to https://github.com/sonic-net/SONiC/pull/2401

PR #252, opened 2026-06-16
[ci] enable PR checker trigger for release branches

yijingyan2:
The Azure Pipelines PR checker for sonic-dash-ha only triggered on PRs targeting the master branch. PRs raised against release branches (e.g. 202605) did not run the PR validation pipeline, leaving release-branch changes unverified.
#### How did you do it?
Added the 202??? branch pattern to the pr.branches.include list so the PR checker also triggers for release branches.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,12 +5,14 @@
 
 pr:
 - master
+- 202???
 
 trigger:
   batch: true
   branches:
     include:
     - master
+    - 202???
 
 pool:
   vmImage: ubuntu-latest


yxieca (approved):
Correct and complete. Adding `202???` to both `pr:` and `trigger:` validates release-branch PRs and builds — the 202605 branch exists here, so these were previously unverified.

Checked:
- The `202???` glob is right — `?` matches a single character in Azure DevOps branch filters, so it matches `202605`, `202505`, etc., consistent with the rest of sonic-net.
- Good that both `pr:` and `trigger:` were updated.
- Azure pipeline is green (amd64-bookworm and amd64-trixie).

One small thing: the PR description says "The Azure Pipelines PR checker for **sonic-dash-ha**…" — looks like a copy-paste from the sibling PR; this repo is sonic-platform-vpp. Worth fixing the body, not blocking.

LGTM.

PR #260, opened 2026-07-10
Fix docker-sonic-vpp single-container image build and runtime

mmiklus:
The single-container target (docker-sonic-vpp.gz) had rotted after the combined-libsai refactor and the rules.mk relocation, and its "not tested / could be broken" README note masked a chain of breakages. Restore it so it builds, starts, and runs a live VPP data plane end to end.

- rules.mk / rules.dep: re-add the dropped docker-sonic-vpp.mk (+ .dep) include that left $(DOCKER_SONIC_VPP) undefined, and wire in vppcfgd.mk (+ .dep) so the vppcfgd daemon is actually built.
- rules/vppcfgd.mk: build the wheel from the in-repo overlay source ($(PLATFORM_PATH)/platform/mkrules/src/sonic-vppcfgd) instead of sonic-buildimage's src/, so no copy into buildimage is needed (sonic-vppcfgd is not an upstream buildimage package).
- rules/docker-sonic-vpp.mk: SYNCD_VPP -> SYNCD_VS and LIBSAIVPP_DBG / SYNCD_VPP_DBG -> LIBSAIVS_DBG / SYNCD_VS_DBG. Those variables were renamed by the combined vs/vpp libsai refactor; the stale names expanded empty, so the image shipped with no VPP/syncd/libsai binaries at all.
- rules/vppcfgd.dep: rewrite to the standard wheel dep pattern (it referenced nonexistent sonic_vppcfgd.mk/.dep files).
- docker-sonic-vpp/Dockerfile.j2: remove the obsolete apt-key + jessie sonic-dev repo (apt-key is gone on bookworm/trixie); purge the Debian python3-cffi so pip can manage cffi for the wheels; drop the conflicting scapy pip pin (scapy comes from the python3-scapy apt package); add meson-python/meson/ninja/patchelf for the dbus-python source build; and apt-install python3-gi/python3-dbus/python3-systemd so PyGObject et al are not rebuilt from source.
- docker-sonic-vpp/orchagent.sh: create /zmq_swss before launching orchagent. P4Orch unconditionally binds a ZMQ server on ipc:///zmq_swss/...; the VM image provides this dir as a host bind-mount, but the single container must create it itself or orchagent aborts (zmq_bind ENOENT).
- README.md: correct the single-container build steps (add cd, code block, document the currently-required INCLUDE_FIPS=n and BUILD_SKIP_TEST=y flags and the iptables modprobe prerequisite) and drop the stale "broken / use prior-to-build-change" note.

Note: this needs one companion change in sonic-buildimage (a separate repo): a files/build/versions/dockers/docker-sonic-vpp/versions-py3 pip constraints file for the container.

Verified: docker-sonic-vpp.gz builds (INCLUDE_FIPS=n BUILD_SKIP_TEST=y), loads, and brings up syncd/orchagent/vppcfgd/vpp_init with a live VPP data plane (Ethernet0/1 Linux-CP taps up, ports oper_status up).


mmiklus:
Related sonic-buildimage change: https://github.com/sonic-net/sonic-buildimage/pull/28362

Build result output (test): https://github.com/users/mmiklus/packages/container/package/docker-sonic-vpp

yue-fred-gao (approved):

yue-fred-gao:
Thanks for fixing the docker build. What is vppcfgd and how is it used?

mmiklus:
> Thanks for fixing the docker build. What is vppcfgd and how is it used?

vppcfgd uses the [vpp_template.json](https://github.com/sonic-net/sonic-platform-vpp/blob/master/docker-sonic-vpp/conf/vpp_template.json) to create vpp startup config - https://github.com/sonic-net/sonic-platform-vpp/blob/master/platform/mkrules/src/sonic-vppcfgd/vppcfgd/vars.py#L20

PR #257, opened 2026-07-02
Update sonic-ext plugin to support portchannel punt 

yue-fred-gao:
### why 
The current port-channel implementation punts control-plane packets directly to the `PortChannel<id>` (bond) netdev. 
This does not conform to the SONiC data-plane model. On real hardware the NPU punts control traffic to the CPU on the physical/member port it was received on. 
The most visible symptom is **DHCP relay**. `sonic-dhcp-relay` / `dhcrelay`
(the SONiC relay agent) binds and accepts uplink packets **only on port-channel member interfaces**. Because today's solution delivers the frame on the `PortChannel<id>` netdev instead of the member, the relay agent never sees the
uplink packet and DHCP relay over a port-channel uplink is broken.

### what this PR does
Update sonic-ext plugin to recognize bond and bond sub interface as aggregate interface to apply punt-via-member behavior.

#### Back port request
- [x] 202605

#### Tested branch
- [x] 202605

#### Test result
- Validation PR: https://github.com/sonic-net/sonic-buildimage/pull/29067 at `2e297e93668c2a791fba6a590befa8eb15bb5ca9`
- Azure build: `1196355`
- Environment: public sonic-mgmt `202605`, `t0-vpp`, one KVM
- TestPlan: https://elastictest.org/scheduler/testplan/6a84bdd4e437251d82f52a1a
- Result: `dhcp_relay/test_dhcp_relay.py::test_dhcp_relay_default[isc-relay-agent]` passed. Module summary: 14 passed, 2 skipped; pretest: 9 passed, 3 skipped; posttest: 5 passed.

PR #267, opened 2026-07-30
[sonic_ext] Add glean-redirect node: punt unresolved-nexthop transit to the kernel

lunyue-ms:
### Why I did it

When a transit next hop is unresolved, VPP generates ARP itself. The reply reaches Linux unsolicited and can be discarded, so `neighsyncd` never learns the neighbor and SAI never programs it back into VPP. Hardware platforms instead punt the original packet to Linux, allowing the kernel to own neighbor resolution.

This change restores that behavior for sonic-platform-vpp and fixes a device-input feature issue exposed by LAG sub-interfaces.

### How I did it

- Add `sonic-ext-glean-redirect` to the IPv4 and IPv6 drop arcs. It recognizes packets produced by the glean/ARP nodes, restores ingress and VLAN metadata, and redirects a throttled copy through the ingress LCP host tap.
- Do not toggle device-input features through sub-interfaces. Their feature configuration is shared with the parent, so sub-interface teardown could remove `sonic-ext-host-xc` from a LAG member.
- Expose glean redirect state and counters through `show sonic-ext`.
- Rebase onto master after #257 merged. The bond/aggregate helpers and behavior from #257 are retained; duplicate implementations previously carried by this PR are removed.
- Bump VPP to `2606-0.5`. Published `2606-0.4` packages did not contain this node, causing downstream builds to download a stale plugin.

### How to verify it

Validated with sonic-net/sonic-sairedis#1907 on `vms-kvm-vpp-t1-lag`:

- `docker-syncd-vpp.gz` built successfully from the rebased sources.
- DUT ran VPP `v2606-0.5+b1sonic1`; the glean node was active and `show sonic-ext` exposed its counter.
- Targeted LAG routing and balancing cases passed.
- Full `test_sub_port_interfaces.py`: **29 passed, 3 skipped**. Two teardown errors were only existing VPP loganalyzer messages during SVI cleanup; all dataplane cases, YANG validation, core/config checks, and post-test sanity checks passed.
- All 8 PortChannels and 24 BGP neighbors remained up.

### Notes for reviewers

- Depends on sonic-net/sonic-sairedis#1907.
- Single commit directly on current master.

### DIFF
--- a/rules/vpp.mk
+++ b/rules/vpp.mk
@@ -7,7 +7,7 @@ VPP_VERSION_BASE = 2606
 # https://packages.buildkite.com/sonic-vpp/vpp; if the suffix isn't bumped,
 # downstream sonic-buildimage builds will silently pull stale debs that
 # pre-date the new patch series and end up with VPP/SAI CRC drift.
-VPP_VERSION = $(VPP_VERSION_BASE)-0.5
+VPP_VERSION = $(VPP_VERSION_BASE)-0.4
 VPP_VERSION_SONIC = $(VPP_VERSION)+b1sonic1
 VPP_SRC_PATH = platform/vpp/vppbld
 
--- a/rules/vpp.mk
+++ b/rules/vpp.mk
@@ -7,7 +7,7 @@ VPP_VERSION_BASE = 2606
 # https://packages.buildkite.com/sonic-vpp/vpp; if the suffix isn't bumped,
 # downstream sonic-buildimage builds will silently pull stale debs that
 # pre-date the new patch series and end up with VPP/SAI CRC drift.
-VPP_VERSION = $(VPP_VERSION_BASE)-0.4
+VPP_VERSION = $(VPP_VERSION_BASE)-0.5
 VPP_VERSION_SONIC = $(VPP_VERSION)+b1sonic1
 VPP_SRC_PATH = platform/vpp/vppbld
 
--- a/vppbld/plugins/sonic_ext/glean_redirect_node.c
+++ b/vppbld/plugins/sonic_ext/glean_redirect_node.c
@@ -242,7 +242,6 @@ VLIB_NODE_FN (sonic_ext_glean_redirect_node)
 (vlib_main_t *vm, vlib_node_runtime_t *node, vlib_frame_t *frame)
 {
   sonic_ext_main_t *sem = &sonic_ext_main;
-  vnet_main_t *vnm = vnet_get_main ();
   clib_thread_index_t thread_index = vm->thread_index;
   u32 n_left_from, *from;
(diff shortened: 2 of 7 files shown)


lunyue-ms:
Hi @yue-fred-gao, this PR follows the design principle you raised in #237 — that the VPP datapath should match how other hardware platforms behave. The main change is a new  glean-redirect  node: when a transit packet hits an unresolved next-hop, VPP's own  ip4-glean  sends the ARP request and drops the packet, so the reply arriving at the kernel is unsolicited and gets discarded ( arp_accept=0 ) — the neighbour is never learned and never reaches SAI. The node punts that packet to the ingress host tap instead, so the kernel routes it and sends its own ARP, mirroring what an NPU does on an unresolved next-hop.
Would you mind reviewing whether this fits our design rules? Thanks!

lunyue-ms:
@yue-fred-gao, I rebased the code and retest all the sub_port_interfaces tests. Would you please review the change. Thanks!

vppbld/plugins/sonic_ext/glean_redirect_node.c
```diff
@@ -0,0 +1,486 @@
...
+       * enough to (re)arm the kernel's own ARP; the rest keep
+       * dropping until the neighbour is programmed via SAI. */
+      key = ((u64) adj_index0 << 32) | orig_rx;
+      if (throttle_check (&sonic_ext_glean_throttle, thread_index, key, seed))
+	{
+	  n_throttled++;
+	  goto trace0;
+	}
+
+      swo = vnet_get_sw_interface (vnm, orig_rx);
```
yue-fred-gao: Since below code is mostly identical to sonic-ext-aggr-tap-redirect, can we extract the common code and put it in a function in sonic_ext.c? I suspect this may be used in some other place in the future.

rules/vpp.mk
```diff
@@ -7,7 +7,7 @@ VPP_VERSION_BASE = 2606
 # https://packages.buildkite.com/sonic-vpp/vpp; if the suffix isn't bumped,
 # downstream sonic-buildimage builds will silently pull stale debs that
 # pre-date the new patch series and end up with VPP/SAI CRC drift.
-VPP_VERSION = $(VPP_VERSION_BASE)-0.4
+VPP_VERSION = $(VPP_VERSION_BASE)-0.5
```
yue-fred-gao: Please remove the version change. We will update the version in the next monthly vpp release.

vppbld/plugins/sonic_ext/glean_redirect_node.c
lunyue-ms: Good point! Will refactor the code. Thanks!

yue-fred-gao (approved):

PR #269, opened 2026-08-03
[vpp] Add sonic_ext packet-trim admission shim (SAI DROP_AND_TRIM datapath)

aaronber0614:
### What / Why

Adds the `sonic_ext` VPP plugin datapath that implements the SAI
`DROP_AND_TRIM` packet-trimming behavior for the SONiC-VPP platform.

Because SONiC-VPP has no hardware queue-admission engine, trimming is
implemented as a software **admission shim** on the `interface-output`
feature arc. When a trim-eligible egress queue cannot admit a packet, the shim
truncates it to the configured size, rewrites DSCP, and re-queues it on the
configured static trim queue on the same physical egress; non-eligible or
unconfigured queues are passed straight through unpoliced (forwarding
unchanged).

Part of the SONiC packet-trimming enablement tracked in
sonic-net/sonic-buildimage#25789.

### Changes

- New `sonic_ext` plugin sources: `trim_admission_node.c`, `trim_node.c`,
  `trim_api.c`, `trim_cli.c`, `sonic_ext_trim.api`, and `sonic_ext.h` state.
- Binary API + debug CLI to program global trim policy and per-`{port,queue}`
  admission/eligibility state.

### Disabling / opting out

The feature is **opt-in and off by default**, and there is no way for it to
perturb normal forwarding unless explicitly configured. There are four
practical levels of control:

1. **Default off (no configuration).** `sonic-ext-trim-admission` is *registered*
   on the `interface-output` arc but is not enabled on any interface at init.
   Until a queue is bound to a `DROP_AND_TRIM` buffer profile, the arc is
   enabled on no port and there is **zero incremental trim datapath cost**.
2. **Per-port / per-queue eligibility — the real teardown path.** The admission
   arc is enabled on a port only while that port has ≥1 trim-eligible queue
   (buffer profile == `DROP_AND_TRIM`). Removing eligibility from a port's last
   eligible queue (unbind/change that queue's buffer-profile binding via SAI
   SET) tears the arc back down and returns the port to zero incremental cost.
   *Ordering caveat:* this teardown runs only while trimming is still globally
   enabled (the SAI refresh path early-returns once global trim is off), so
   remove the buffer profiles **before** the global disable to fully return to
   zero cost.
3. **Global functional off switch.** `SAI_SWITCH_ATTR_PACKET_TRIM_SIZE = 0`
   (equivalently the plugin `sonic_ext_trim_global_set(is_enable=false)` or the
   debug CLI `disable`) clears global policy so **no packet is trimmed
   switch-wide**. This is a *functional* off switch — the admission node
   bypasses every packet at its first check. By itself it does **not** tear down
   already-enabled per-port arcs, so a port that previously had an eligible
   queue keeps the arc installed and each egress packet still takes a cheap
   admission-node dispatch that immediately bypasses. It is a no-op, not
   literally zero cost; combine with step 2 (in that order) to also drop the
   arcs.
4. **Coarse whole-plugin disable.** VPP's generic plugin loader honors
   `plugin sonic_ext_plugin.so { disable }` in `startup.conf`, which removes the
   entire `sonic_ext` plugin (and therefore all trim behavior). It is a blunt
   instrument — it also disables the plugin's other functions — not a
   trim-specific knob.

There is **no** packet-trim-specific compile-time flag; the trim sources are
compiled unconditionally into `sonic_ext_plugin.so`.

### Testing

- VPP framework plugin tests for the admission/trim nodes.
- End-to-end PTF packet-trimming stages 1–2 pass on a `t1-lag-vpp` testbed
  (paired with the companion sonic-sairedis and sonic-mgmt changes).

### Deferred (by design, tracked in the HLD)

Asymmetric `FROM_TC` DSCP, dynamic queue resolution, ACL `DISABLE_TRIM`,
per-port/queue SAI counter sourcing, and multi-worker bucket sharding.

> **Merge ordering:** land this VPP plugin PR first — it publishes the
> `sonic_ext_trim.api` headers the companion sonic-sairedis translation PR
> compiles against.

---
🤖 Co-authored with GitHub Copilot.

### DIFF
--- a/rules/vpp.mk
+++ b/rules/vpp.mk
@@ -8,7 +8,7 @@ VPP_VERSION_BASE = 2606
 # https://packages.buildkite.com/sonic-vpp/vpp; if the suffix isn't bumped,
 # downstream sonic-buildimage builds will silently pull stale debs that
 # pre-date the new patch series / plugins and end up with VPP/SAI CRC drift.
-VPP_VERSION = $(VPP_VERSION_BASE)-0.7
+VPP_VERSION = $(VPP_VERSION_BASE)-0.8
 VPP_VERSION_SONIC = $(VPP_VERSION)+b1sonic1
 VPP_SRC_PATH = platform/vpp/vppbld
 
--- a/rules/vpp.mk
+++ b/rules/vpp.mk
@@ -1,13 +1,14 @@
 # libvpp package
 
 VPP_VERSION_BASE = 2606
-# Bump the minor suffix whenever vppbld/patches/series or any patch file
-# under vppbld/patches/*.patch changes content. The VPP_VERSION_SONIC string
-# is the cache key used by vppbld/Makefile to fetch pre-built debs from
+# Bump the minor suffix whenever vppbld/patches/series, any patch file
+# under vppbld/patches/*.patch, or any custom plugin under vppbld/plugins/
+# changes content. The VPP_VERSION_SONIC string is the cache key used by
+# vppbld/Makefile to fetch pre-built debs from
 # https://packages.buildkite.com/sonic-vpp/vpp; if the suffix isn't bumped,
 # downstream sonic-buildimage builds will silently pull stale debs that
-# pre-date the new patch series and end up with VPP/SAI CRC drift.
-VPP_VERSION = $(VPP_VERSION_BASE)-0.3
+# pre-date the new patch series / plugins and end up with VPP/SAI CRC drift.
+VPP_VERSION = $(VPP_VERSION_BASE)-0.6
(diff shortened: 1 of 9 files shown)


aaronber0614:
**Companion PRs** (SONiC-VPP packet trimming, tracked in sonic-net/sonic-buildimage#25789):
- HLD: sonic-net/SONiC#2483
- VPP `sonic_ext` plugin datapath: sonic-net/sonic-platform-vpp#269
- SAI translation: sonic-net/sonic-sairedis#2017
- Test enablement: sonic-net/sonic-mgmt#26743

Suggested review/merge order: HLD → sonic-platform-vpp → sonic-sairedis → sonic-mgmt.

PR #272, opened 2026-08-09
[sonic_ext] Add ip2me receive-DPO ACL-bypass feature node

### LINKED ISSUE #28884: Bug: [vpp][dualtor] Mux drop ACL drops control-plane/for-us traffic — IN_PORTS ignored and L2 ACL runs before ip4-local punt (mux stuck unhealthy)
### Is it platform specific
generic (VPP platform / `docker-syncd-vpp`, `libsaivs` vslib + VPP graph)

### Importance or Severity
High

### Description of the bug
On the VPP platform, the dual-ToR **mux drop ACL** (`INGRESS_TABLE_DROP` / `mux_acl_rule`, MuxOrch) drops **control-plane / for-us (IP2ME)** traffic that should be trapped to the CPU. On a hardware ASIC these packets (ARP, ICMP-to-me, BGP, mux ICMP heartbeat) are trapped by the my-station / CoPP path **before / independently of** the ingress data ACL, so a standby mux port still delivers them to the control plane; the data-drop ACL only removes the *forwarding* copy. VPP does not preserve this ordering.

Two distinct VPP fidelity gaps combine here:

**(1) VPP ignores `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS`.**
MuxOrch models the drop as one ACL table bound to *all* mux ports, with a per-rule `MATCH_IN_PORTS = {standby ports}` qualifier that scopes the DROP to standby ports only. In `vslib/vpp/SwitchVppAcl.cpp`, `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` is never read. The table→port binding is honored, but the `IN_PORTS` qualifier is silently dropped, so the rule degrades to a wildcard `ipv4 deny src 0/0 dst 0/0` applied to every bound mux port — **including active ports**. Editing `IN_PORTS` on switchover has no effect in VPP.

**(2) The mux drop ACL runs at `l2-input`, before the for-us / L3 punt at `ip4-local`.**
VPP realizes the mux ACL as an L2-stage ACL (`acl-plugin-in-ip4-l2`) at `l2-input` — the first stage after the frame enters the bridge domain. The for-us recognition and punt happen much later, at `ip4-input on BVI → ip4-lookup → ip4-local / ip4-punt` (VLAN-BVI HLD §5.6). So a for-us packet is denied at L2 before it can ever be recognized as IP2ME and punted. Even a correctly-scoped **standby** port would still drop its control-plane/heartbeat traffic, because there is no "trap for-us/CoPP before the data ACL" semantics in the VPP pipeline.

Concrete impact: the linkmgrd mux ICMP heartbeat reply (server → ToR loopback, a for-us packet) is dropped on the mux port, so linkmgrd never confirms the port and it is stuck `HEALTH=unhealthy`. Because the drop also eats the heartbeat that would clear the condition, an active port that still has the (mis-scoped) deny bound never recovers.

### Steps to Reproduce
1. Dual-ToR VPP KVM testbed (`subtype=DualToR`), mux ports on `Vlan1000`.
2. Observe `show mux status` — a port reports `active` but `HEALTH=unhealthy`.
3. On VPP:
   - `vppctl show acl-plugin acl` → a `ipv4 deny src 0.0.0.0/0 dst 0.0.0.0/0` rule (tag `sonic_acl_oid:...`) applied inbound on the mux bridge-member sw_if_indexes.
   - `vppctl show errors` → `acl-plugin-in-ip4-l2 ACL deny packets` incrementing in step with the port's ingress.
   - `vppctl show ip fib <ToR loopback>` → `dpo-receive on loopN` (for-us), i.e. it *would* be punted at `ip4-local` — after the L2 ACL that already dropped it.
4. `ASIC_DB` shows the active port's OID **removed** from `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS`, yet VPP still has the deny ACL bound to that port.

### Actual Behavior and Expected Behavior
**Actual:** Mux drop ACL is applied as an unscoped `deny any/any` at `l2-input` on every bound mux port; `IN_PORTS` is ignored, and control-plane/for-us traffic (ARP/ICMP-to-me/BGP/heartbeat) is dropped before the L3 for-us punt. Active ports drop too. Mux health cannot recover.

**Expected (match hardware semantics):**
- Honor `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` so the DROP is scoped to the standby ports actually in the set (bind/unbind the VPP ACL per member, re-derived on `set`), so active ports never drop.
- Control-plane / for-us (IP2ME) traffic must be trapped/punted **before** the mux data-drop ACL consumes it (as on hardware, where CoPP/my-station trapping precedes the ingress data ACL). The mux drop should only remove the forwarding copy, never the CPU-punt copy.

### Relevant log output
```
# mux says active+unhealthy:
show mux status | grep Ethernet4
  Ethernet4  active  active  unhealthy  consistent  ...

# ASIC_DB removed the active port from IN_PORTS (correct intent):
redis-cli -n 1 hget ASIC_STATE:SAI_OBJECT_TYPE_ACL_ENTRY:oid:0x8000000000692 SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS
  -> Ethernet4 oid 0x1000000000033 ABSENT

# ... but VPP still has the deny ACL bound to that port (sw_if 2 = bobm1 = Ethernet4):
vppctl show acl-plugin interface sw_if_index 2   -> input acl(s): 2, 3, 1   (3 = deny any/any)

# ... dropped at L2, before the for-us punt:
vppctl show errors | grep 'acl-plugin-in-ip4-l2'
  478550  acl-plugin-in-ip4-l2   ACL deny packets
vppctl show ip fib 10.1.0.36
  [@0]: dpo-receive: 10.1.0.36 on loop2      # for-us; would punt at ip4-local (after the L2 ACL)
```

### Output of `show version`, `show techsupport`
```
SONiC Software Version: SONiC.master.1182394-fc9ea17bb
Distribution: Debian 13.5
Platform: x86_64-kvm_x86_64-r0
HwSKU: Force10-S6000
ASIC: vpp
ASIC Count: 1
```

### Fix direction
- `vslib/vpp/SwitchVppAcl.cpp`: implement `SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS` by binding/unbinding the VPP ACL to exactly the member interfaces in the set, re-derived on entry `set` (so removing a port from `IN_PORTS` unbinds the deny from that port).
- Ensure control-plane / for-us (IP2ME) punt precedes the mux data-drop ACL — e.g. a higher-priority `l2-input-classify` punt arc for for-us / control ethertypes+L4 (analogous to the existing LLDP/DHCP punt arcs in the `sonic_ext` plugin), so the CPU-punt copy survives the mux drop. See VPP VLAN-BVI HLD §5.6 (`doc/vpp/vlan-bvi-hld.md`); the HLD does not currently cover the ACL/mux/CoPP ordering.

Related: sonic-net/sonic-buildimage#28882 (VLAN SVI neighbor host route not installed) — separate VPP dual-ToR data-path gap.

lolyu:
This fixes: https://github.com/sonic-net/sonic-buildimage/issues/28884

Add sonic-ext-ip2me-ip4/ip6 nodes on the l2-input-ip4/ip6 feature arcs, running before acl-plugin-in-ip*-l2. Each frame gets a FIB receive-lookup on the inner dst IP; a DPO_RECEIVE (ip2me) result steers it to l2-input-feat-arc-end, bypassing the ACL so for-us traffic still reaches the BVI/ip*-local for punt while transit traffic hits the ACL as usual. This restores hardware-style CoPP-before-ACL ordering so an ingress drop ACL (e.g. the dual-ToR mux drop, which degrades to a port-wide deny 0/0 under VPP) no longer discards control traffic like the ICMP mux heartbeat. Zero ACL rules consumed; no per-address state.

Enabled per-interface via a new sonic_ext_ip2me_enable_disable binary API, with a "sonic-ext ip2me <intf> [on|off]" CLI and an "ip2me hits" counter.

* work item ID: 39250596


lolyu:
Hi @yue-fred-gao, this PR is the second approach to fix the "ACL dropping for-us packets" issue. I think this is a cleaner approach to fix this issue as it emulates the hardware behavior of CoPP before ACL in real hardware.

vppbld/plugins/sonic_ext/sonic_ext.h
```diff
@@ -135,6 +141,14 @@ void sonic_ext_host_xc_enable_disable (u32 sw_if_index, int enable);
  * tomorrow).  Driven from the LCP pair add/del callback. */
 void sonic_ext_aggr_tap_redirect_enable_disable (u32 sw_if_index, int enable);
 
+/* Enable / disable the "receive-DPO check before ACL" feature on a
+ * given L2 port sw_if_index.  Registers sonic-ext-ip2me-ip4/ip6 on the
+ * l2-input-ip4 / l2-input-ip6 feature arcs, ahead of the ACL plugin, so
+ * that ip2me (destined-to-this-switch) traffic bypasses any ingress drop
+ * ACL that would otherwise discard it.  Driven from the SAI-VPP layer
+ * wherever such a drop ACL is bound. */
```
yue-fred-gao, line 166: Is it easy to identify such drop ACL at sai layer?

yue-fred-gao (commented):
I like this approach more than the ACL permit option. I think it is faster than programming permit rules for each for-me ip.

vppbld/plugins/sonic_ext/sonic_ext.h
lolyu: Yes — it's a cheap, exact check rather than a heuristic, because we don't try to recognize a specific named ACL (e.g. "the mux drop ACL"). We just flag any ingress ACL table that contains at least one DROP rule, which is deliberately generic so it covers the mux drop ACL today and any future drop ACL that could discard ip2me traffic.

yue-fred-gao (approved):

PR #273, opened 2026-08-13
Update mirror HLD with deferred mirror for egress mirroring

yue-fred-gao:
### why
The vpp-mirror HLD doesn't support ingress acl with egress mirroring action in everflow. This requires redesign the mirroring action in ACL and a new node in interface-output arc.

### what this PR does
Update mirroring action in ACL
Add sonic-ext-egress-mirror node to interface-output arc.


yue-fred-gao:
The HLD has been moved to SONiC

PR #275, opened 2026-08-20
fix(vpp): backport portchannel punt support to 202605

auspham:
## Summary

Backport the PortChannel punt-via-member support from upstream PR #257 to the `202605` release branch.

Original PR: https://github.com/sonic-net/sonic-platform-vpp/pull/257

## Conflict resolution

The `202605` branch diverged before PR #255 introduced the `sonic_ext` plugin, so the three files modified by #257 were absent rather than relocated. The active custom-plugin source path remains `vppbld/plugins`, which `vppbld/Makefile` copies into the VPP source tree during a local build.

Instead of restoring only the three conflicted fragments, this backport adds the complete `sonic_ext` plugin tree exactly as it exists at the #257 merge commit and enables it in both VPP startup templates. This supplies the prerequisite plugin baseline and the PortChannel-specific behavior at the branch's active implementation path.

## Validation

- `git diff --check`
- Confirmed no unresolved conflict markers
- Confirmed every added `sonic_ext` file is byte-for-byte identical to merge commit `53a0712b199b5cba54b6a433ca0b8c688d2c60cc`
- `make -s -n -C vppbld repo_clone`
- Existing 202605 functional validation: https://github.com/sonic-net/sonic-buildimage/pull/29067

### DIFF
--- a/docker-sonic-vpp/conf/startup.conf.tmpl
+++ b/docker-sonic-vpp/conf/startup.conf.tmpl
@@ -299,8 +299,3 @@ plugins {
    ## Set per-class configuration
    # class dpdk/cryptodev { rate-limit 100 level debug syslog-level error }
 # }
-
-# Why not support sub interfaces
-linux-cp {
-   lcp-auto-subint
-}
--- a/docker-sonic-vpp/conf/startup.conf.tmpl
+++ b/docker-sonic-vpp/conf/startup.conf.tmpl
@@ -252,6 +252,7 @@ plugins {
         plugin vxlan_plugin.so { enable }
 	plugin tunterm_acl_plugin.so { enable }
 	plugin ip_validate_plugin.so { enable }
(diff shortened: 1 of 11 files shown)


auspham:
Added the missing PR #255 startup prerequisite in commit `1b24e83`: disable `linux-cp lcp-auto-subint` in both VPP startup templates. The VLAN-BVI HLD requires this to keep tagged control-plane traffic on the physical member netdev where DHCP relay listens; leaving it enabled can create `Ethernet<n>.<vid>` LCP pairs that intercept the punted frame before VLAN delivery.

With this change plus #276, the `sonic_ext` plugin sources and `VPP_VERSION=2606-0.4` are identical to the platform tree used by successful 202605 validation build `1196355` / Elastictest plan https://elastictest.org/scheduler/testplan/6a84bdd4e437251d82f52a1a.

PR #219, opened 2026-04-07
Support test_drop_counters.py in Sonic-VPP

AkeelAli:
Changes to support passing the tests in test_drop_counters.py in Sonic-VPP

Changes include:
- New VPP plugin `ip_validate_plugin` to drop invalid packets not natively dropped by VPP. Once merged, the plan is to try to upstream it to fd.io.
- VPP patch to count bond drops against the member interface (to show up under the member's RX_DRP in sonic)

Associated sonic-mgmt PR: https://github.com/sonic-net/sonic-mgmt/pull/24031

46 new sonic-vpp passing tests:

| Test Name | Extra Variations (+rif,pc) | Plugin Required | Skip Counter | Total |
|---|---|---|---|---|
| test_dst_ip_is_loopback_addr | — | Yes | - | 2 |
| test_src_ip_is_loopback_addr | — | Yes | - | 2 |
| test_src_ip_is_multicast_addr | IPv4, IPv6 | Yes | - | 4 |
| test_src_ip_is_class_e | — | Yes | - | 2 |
| test_ip_is_zero_addr | IPv4/IPv6 × src/dst | Yes | - | 8 |
| test_dst_ip_link_local | — | Yes | - | 2 |
| test_src_ip_link_local | — | Yes | - | 2 |
| test_unicast_ip_incorrect_eth_dst | 2 MAC variants | Yes | - | 4 |
| test_dst_ip_absent | — | Yes | - | 2 |
| **SUBTOTAL** | | | | **28** |
| test_not_expected_vlan_tag_drop | — | - | Yes | 2 |
| test_ip_pkt_with_exceeded_mtu | — | - | Yes | 2 |
| test_ip_pkt_with_expired_ttl | — | - | - | 2 |
| test_broken_ip_header | version, chksum, ihl | - | - | 6 |
| test_absent_ip_header | — | - | - | 2 |
| test_acl_drop | — | - | - | 2 |
| test_acl_egress_drop | — | - | - | 2 |
| **SUBTOTAL** | | | | **18** |
| **TOTAL** | | | | **46** |

----------------

VPP changes tested separately too:

TEST RESULTS:
IPv4: unicast IP dst with multicast ETH dst (01:00:5e) -> drop      2.19 OK
IPv6: unicast IP dst with broadcast ETH dst -> drop                 2.19 OK
IPv6: unicast IP dst with multicast ETH dst -> drop                 2.22 OK
IPv4: valid packets traverse node and increment VALID counter       1.21 OK
[...]

PR #279, opened 2026-09-01
Support test_drop_counters.py in t0-vpp

AkeelAli:
### Summary

Changes to support passing test_drop_counters.py in t0-vpp.

Redesigns the previous solution from https://github.com/sonic-net/sonic-platform-vpp/pull/219 by replacing the VPP patch with a plugin-only node to credit drops to the original LAG/VLAN member interface.

## t1-lag-vpp
https://github.com/sonic-net/sonic-buildimage/pull/29463: sonic-buildimage with current platform/vpp changes to ensure test_drop_counters.py continues to pass in t1-lag-vpp and no regressions are introduced in t1-lag-vpp and t0-vpp.

test_drop_counters.py still passing in t1-lag-vpp:
https://elastictest.org/scheduler/publictestplan/6aa5baafc6c96c8b4c45714b?searchTestCase=test_drop&testcase=drop_packets%2Ftest_drop_counters.py&type=console

```
========== 46 passed, 56 skipped, 1353 warnings in 1940.94s (0:32:20) ==========
```

## t0-vpp
https://github.com/sonic-net/sonic-mgmt/pull/27901: sonic-mgmt with image from above sonic-buildimage and test_drop_counters.py enabled in t0-vpp.

test_drop_counters.py passing in t0-vpp with current change:
https://elastictest.org/scheduler/publictestplan/6aa80a162963ddabcd34ba40?searchTestCase=test_drop_counters.py&testcase=drop_packets%2Ftest_drop_counters.py&type=console
```
========== 41 passed, 61 skipped, 1311 warnings in 2881.06s (0:48:01) ==========
```


vppbld/plugins/sonic_ext/drop_member_stats_node.c
```diff
@@ -0,0 +1,203 @@
...
+  /* Catch interfaces that already existed before the add/del hook was
+   * registered (local0 and anything created during early boot). */
+  pool_foreach (si, im->sw_interfaces)
+    {
+      sonic_ext_drop_member_stats_enable (si->sw_if_index);
+    }
+  return 0;
+}
+
+VLIB_MAIN_LOOP_ENTER_FUNCTION (sonic_ext_drop_member_stats_loop_enter);
```
yue-fred-gao, line 203: I think this only enables the feature in the interfaces that are created when the main loop calls this function, which should include all the physical interfaces. We don't need to enable the feature on dynamically created interfaces, such as bvi or lag, right? 
There is no way to disable this feature unless we remove sonic-ext plugin, correct? Can you please look at the option to enable/disable a feature using startup.conf? For applications don't care sonic compliant, they can opt out this feature using the config? That removes the code level of dependency to enable/disable through API. If this works, other features in sonic-ext should follow.

AkeelAli: `VNET_SW_INTERFACE_ADD_DEL_FUNCTION` will enable the feature on dynamically created interfaces including BVI and LAG. It needs to be enabled on all interfaces because of how `interface_drop_punt()` dispatches the error-drop arc for the whole frame using a single interface's feature config (likely a performance decision). See https://github.com/FDio/vpp/blob/0663bc8dcbd08c671131d61c775cf6e9b790418c/src/vnet/interface_output.c#L1050
A single error-drop frame can batch drops from multiple RX interfaces, but only one of them decides whether the whole frame is sent to our feature node. So if the feature isn't enabled everywhere, LAG/BVI drops batched with other interfaces could be missed.

Regarding the enabling/disabling of the sonic-ext features, I think it's a great idea to do it via config as you suggested and I plan to follow-up on this in a subsequent PR. The performance penalty from visiting our node should be limited for now as it is only incurred in the drop path and the node short-circuits on non-LAG/BVI drops.

yue-fred-gao: sounds good

yue-fred-gao (approved):

PR #281, opened 2026-09-03
[vpp] add plugins for CoPP dataplane enforcement

nhegde-microsoft:
## What
CoPP dataplane enforcement, in the shared `sonic_ext` VPP plugin:
- `copp_ifout_node`: ARP/LACP/LLDP/UDLD/TTL_ERROR on `interface-output`.
- `copp_ip2me_node`: IP2ME/SNMP/SSH/BGP/BGPV6 on `ip4-punt`, one independently-keyed policer slot per SAI trap group.
- `copp_udld_node`: UDLD has no EtherType (802.3 length field instead), so it can't be matched by `copp_ifout_node`'s byte read. This node catches it via VPP's real LLC-null/LLC+SNAP+Cisco-UDLD dispatch and pre-tags the buffer with the right `copp_ifout_node` entry.

## Why
[sonic-buildimage#25801](https://github.com/sonic-net/sonic-buildimage/issues/25801). Ethertype-keyed traps never reach VPP's classify arcs on `linux-cp`-paired ports. IP2ME/SNMP/SSH/BGP/BGPV6 have no ethertype to match on. UDLD needs real protocol dispatch, not a byte match.

## How
- `copp_ifout_node` binds only on a bond member's own tap, never the aggregate's (avoids double-policing a member→aggregate→member redirected packet).
- `copp_ip2me_node` is a single global feature on `ip4-punt` (no per-interface binding needed). Meters via the existing SAI-created policer; conform/unmatched → `ip4-punt-redirect`, exceed/violate → `ip4-drop`.
- `copp_udld_node` rewinds to the intact Ethernet header, tags the buffer, and re-enters `copp_ifout_node`'s arc via `vnet_feature_arc_start()`.

Design: https://github.com/nhegde-microsoft/SONiC/blob/copp-vpp-enablement/doc/vpp/vpp_copp_HLD.md

## Companion PRs
- sairedis: https://github.com/sonic-net/sonic-sairedis/pull/2070
- sonic-mgmt: https://github.com/sonic-net/sonic-mgmt/pull/27726
- HLD: https://github.com/sonic-net/SONiC/pull/2539

## Test
`vlab-vpp-01`: all 24 applicable `tests/copp/test_copp.py` subtests pass.


aaronber0614:
## Multi-model review: sonic-platform-vpp #281 (copp_punt_policer plugin)

I ran this through a three model consensus review (Claude Opus, GPT, Grok, all at high reasoning) and then a second adversarial pass whose job was to disprove each finding. Every item below is grounded against the pinned VPP revision `3f9e978d` (policer lives at `src/plugins/policer/`) and the local sonic-sairedis tree. CI on this PR is green and the plugin itself is well structured. The blocker below is a deploy wiring gap, not a defect in the plugin C code.

### Blocker 1: startup.conf.tmpl never enables the native `policer_plugin.so`

`docker-syncd-vpp/conf/startup.conf.tmpl` uses `plugin default { disable }` and then enables an explicit allow list. This PR adds `plugin copp_punt_policer_plugin.so { enable }` but does not add `plugin policer_plugin.so { enable }`. In the pinned VPP, policer is `add_vpp_plugin(policer ...)` in `src/plugins/policer/CMakeLists.txt`, so it is a loadable `policer_plugin.so`, not core vnet. Its binary API messages register only when that plugin loads.

The sonic-sairedis side (#2070) then does, in `get_base_msg_id()`:

```
policer_msg_id_base = vl_client_get_first_plugin_msg_id("policer_%08x...");
assert(policer_msg_id_base != (u16)~0);
```

On a clean build and deploy where `policer_plugin.so` is not loaded, that lookup returns `~0` and the assert aborts syncd at connect time (this aborts all of syncd, not just CoPP). `configure.ac` does not define `NDEBUG`, so the assert is live; even if it were compiled out, a `~0` base makes every `POLICER_ADD` a garbage message id that never reaches the real handler, so no policer is ever programmed.

I confirmed the plugin data path does NOT crash in this scenario: `copp_punt_policer_resolve_index()` null checks `policer_get_main()` and returns `DROP_UNRESOLVED`, so bound ethertypes are dropped rather than dereferencing NULL. The failure is purely the syncd side abort above.

I suspect the manual validation passed because the test box had `policer_plugin.so` enabled out of band. Fix: add `plugin policer_plugin.so { enable }` to `startup.conf.tmpl` (and to the `docker-sonic-vpp` startup template if CoPP is meant to work in the single container image too).

### Blocker 2 (shared with #2070): SAI policer counters read stats that this plugin never increments

`copp_punt_policer_node.c` meters with the raw `vnet_police_packet(policer, 256, ...)`. That call computes color and updates token buckets only. The stats segment counters `/net/policer/{conform,exceed,violate}` are incremented ONLY by the `policer_police()` wrapper (`src/plugins/policer/police_inlines.h`, `vlib_increment_combined_counter`), which this node never calls. sonic-sairedis `getPolicerStats` reads exactly those `/net/policer/*` counters, so `SAI_POLICER_STAT_GREEN/YELLOW/RED_*` return 0 for every CoPP ethertype. The plugin does keep its own per ethertype counters, but it exposes them through a separate API that `getPolicerStats` does not read. `test_policer` does not catch this because it measures PTF rx_pps, not SAI stats. Fix options: have the node call `policer_police()` so the combined counters increment, or point `getPolicerStats` at this plugin's own counters. This ties directly to HLD REQ-5 (see my #2539 comment).

### Non blocking notes

1. Worker safety: `copp_punt_policer_bind` mutates `cpm->entries[]` and `n_entries` from the main thread with no `vlib_worker_thread_barrier_sync`, and the counter flush is non atomic. This is fine on a single worker t1-lag-vpp KVM but will race on a multi worker setup. Worth a comment noting the single worker assumption, or a barrier if multi worker is in scope.
2. The per packet ELOG in the node path is not gated on `VLIB_NODE_FLAG_TRACE`, so it fires unconditionally and wraps the elog ring under load. Consider gating it on the trace flag.
3. Nit: the SAI to VPP policer mode mapping keys the 1R3C (SR_TCM) case on `has_pir`, but RFC2697 uses PBS, not PIR. Harmless for CoPP (CIR/CBS only), but the discriminator is slightly off.

### Confirmed correct (I checked these and they are fine)

The fixed 256 byte metering length matches VPP's `QOS_POLICER_FIXED_PKT_SIZE` pps calibration and is intentional. Mixing `vnet_feature_next` with the static `next_nodes[]` (DROP=0, INTERFACE_OUTPUT=1) is the same idiom used by ip_validate and the VPP sample plugin, so that is correct.

### Merge order

This PR should land first so platform-vpp master republishes the VPP debs and the generated `copp_punt_policer.api_enum.h` header, then #2070 (which currently red builds only because that header is not yet on master), then the sonic-mgmt skip lift. Please address Blocker 1 (and ideally Blocker 2) before merge.

Reviewed by AI agent on behalf of aaronber0614.

nhegde-microsoft:
@aaronber0614 
Thanks — both blockers check out. Digging into why testing didn't catch Blocker 1: my earlier validation ran against the DUT's live VPP, which turned out to be pinned at the old `435fda042` (pre-pluginification), not the actual `3f9e978d7` this PR targets. On the old build policer was still core `vnet`, always loaded — so that gap couldn't have surfaced there. That means the earlier "8/8 passing" evidence doesn't validate against the real pinned VPP; I'll redeploy against `3f9e978d7` and re-run before restating any pass results.

**Blocker 1**: Fixed — added `plugin policer_plugin.so { enable }` to `docker-syncd-vpp/conf/startup.conf.tmpl`. Left `docker-sonic-vpp`'s template alone; its plugin list was already out of sync with `docker-syncd-vpp` before this PR (missing `sflow`/`sonic_ext` too), so bringing just the policer line in line there felt like a separate pre-existing issue — happy to open a follow-up if that image matters for CoPP.

**Blocker 2**: Fixed — the node now calls `policer_get_counters()` + `vlib_increment_combined_counter()` right after computing the verdict, using the same 256-byte metered length, so the stats-segment counters `getPolicerStats()` reads are no longer stuck at zero.

**Non-blocking notes**: addressed all three — ELOG now gated on `VLIB_NODE_FLAG_TRACE`, a short comment documenting the single-writer (syncd, main thread) assumption instead of an unverified barrier, and the SR_TCM/PIR-vs-PBS discriminator nit is noted for a follow-up.

Will hold off claiming this is validated again until I've re-run the full suite against a real `3f9e978d7` build.

aaronber0614 (changes requested):
BGPv6 is claimed and bound but not policed

sonic-net/sonic-sairedis#2070 maps both BGP and BGPv6 to the TCP/179 binding, but sonic-net/sonic-platform-vpp#281:

Parses  ip4_header_t.
Registers only on  ip4-punt .
Has no  ip6-punt  counterpart.
Tests only IPv4 BGP traffic.

IPv6 BGP therefore bypasses this CoPP policer. Add an IPv6 punt path and test, or explicitly remove BGPv6 from the supported scope and HLD.

_Co-authored with CoPilot_

nhegde-microsoft:
> BGPv6 is claimed and bound but not policed
> 
> [sonic-net/sonic-sairedis#2070](https://github.com/sonic-net/sonic-sairedis/pull/2070) maps both BGP and BGPv6 to the TCP/179 binding, but #281:
> 
> Parses  ip4_header_t. Registers only on  ip4-punt . Has no  ip6-punt  counterpart. Tests only IPv4 BGP traffic.
> 
> IPv6 BGP therefore bypasses this CoPP policer. Add an IPv6 punt path and test, or explicitly remove BGPv6 from the supported scope and HLD.
> 
> _Co-authored with CoPilot_

Thanks @aaronber0614, good catch!
Looking into how this was missed, turns out test_copp.py (run for all ASICs) never actually tests BGPv6. Since fixing the test affects all ASICs, not just VPP, I'll fix this for VPP and add the test in a follow-on PR. Opened [#29662](https://github.com/sonic-net/sonic-buildimage/issues/29662) to track it.

aaronber0614 (approved):
LGTM

nhegde-microsoft:
> > BGPv6 is claimed and bound but not policed
> > [sonic-net/sonic-sairedis#2070](https://github.com/sonic-net/sonic-sairedis/pull/2070) maps both BGP and BGPv6 to the TCP/179 binding, but #281:
> > Parses  ip4_header_t. Registers only on  ip4-punt . Has no  ip6-punt  counterpart. Tests only IPv4 BGP traffic.
> > IPv6 BGP therefore bypasses this CoPP policer. Add an IPv6 punt path and test, or explicitly remove BGPv6 from the supported scope and HLD.
> > _Co-authored with CoPilot_
> 
> Thanks @aaronber0614, good catch! Looking into how this was missed, turns out test_copp.py (run for all ASICs) never actually tests BGPv6. Since fixing the test affects all ASICs, not just VPP, I'll fix this for VPP and add the test in a follow-on PR. Opened [#29662](https://github.com/sonic-net/sonic-buildimage/issues/29662) to track it.

@aaronber0614 slight change - the BGPv6 code was not difficult to add, so I went ahead and added it. I did not add/enable the test though since it affects all ASICs. I modified [#29662](https://github.com/sonic-net/sonic-buildimage/issues/29662) to do that instead so that whenever we add those, sonic-plaform-vpp will support it.

vppbld/plugins/sonic_ext/capture_node.c
```diff
@@ -112,6 +112,9 @@ VLIB_NODE_FN (sonic_ext_capture_node)
       }
 
       seb->magic = SONIC_EXT_BUFFER_MAGIC;
+
+      /* sonic-ext-copp-udld sets this to a real index for UDLD traffic */
+      seb->copp_ifout_entry_idx = ~0;
```
yue-fred-gao, line 117: The vnet buffer has very limited space. If fd.io adds a new field, we might not be able to fit our extensions. We are just borrowing the unused space today. We should avoid using the vnet buffer if we can. For this particular case, is it possible to detect the UDLD packet at the copp_ifout_node? Something like below?
/* ethertype < 0x0600 => 802.3 length field, so an LLC frame follows */
if (PREDICT_FALSE (ethertype < 0x0600))
  {
    llc_header_t *llc = (llc_header_t *) (eth + 1);

    if (b->current_length >= sizeof (*eth) + sizeof (*llc) + sizeof (snap_header_t) &&
        llc->dst_sap == 0xaa && llc->src_sap == 0xaa && llc->control == 0x03)
      {
        snap_header_t *snap = (snap_header_t *) (llc + 1);

        if (snap_header_get_oui (snap) == IEEE_OUI_cisco &&
            clib_net_to_host_u16 (snap->protocol) ==
              SNAP_cisco_unidirectional_link_detection)
          idx = sonic_ext_copp_ifout_find_entry (sem, SONIC_EXT_COPP_UDLD_ETHERTYPE);
      }
  }

vppbld/plugins/sonic_ext/sonic_ext.api
```diff
@@ -72,3 +72,155 @@ autoreply define sonic_ext_mirror_encap_fixup_enable_disable
   u8 hop_limit;
   bool enable;
 };
+
```
yue-fred-gao, line 75: Please add an API or config to disable/enable this feature. We plan to use startup config to enable/disable features in the sonic-ext plugin: https://github.com/sonic-net/sonic-platform-vpp/pull/288. For applications don't care the feature, they can disable it through config and no extra cost added to data path.

vppbld/plugins/sonic_ext/copp_ip2me_node.c
```diff
@@ -0,0 +1,754 @@
...
+/*
+ * Resolve a policer slot's VPP policer_index by name, lazily -- same
+ * pattern as sonic-ext-copp-ifout, so bind order relative to
+ * policer_add() doesn't matter and a later policer_update() recreating
+ * the object under the same name is picked up automatically.
+ */
+static_always_inline u32
+sonic_ext_copp_ip2me_resolve_index (sonic_ext_copp_ip2me_policer_t *pol)
+{
+  policer_main_t *pm = policer_get_main ();
```
yue-fred-gao, line 94: This is an expensive operation to here:  https://github.com/FDio/vpp/blob/f3aee87367dd11525cc6108841c07b4a321ed9db/src/plugins/policer/policer.h#L142.

policer_get_main() and policer_get_counters() are called per packet, and each one
   performs a dlsym().

   In this VPP tree these are non-internal inlines that do a string hash lookup on
   plugin_by_name_hash followed by dlsym() on every call (src/plugins/policer/policer.h
   lines 138-160, src/vlib/unix/plugin.c line 30). All three new nodes call
   policer_get_main() twice and policer_get_counters() once per packet, inside
   ..._resolve_index() and the metering block. That is 2-3 string hashes plus dlsym per
   punted packet. Resolve both pointers once into file-static caches, at VLIB_INIT or on
   first use, and reuse them.

vppbld/plugins/sonic_ext/copp_ifout_node.c
```diff
@@ -0,0 +1,406 @@
...
+			   int is_bind, int match_ip4_ttl_expiring)
+{
+  sonic_ext_main_t *sem = &sonic_ext_main;
+  int idx = sonic_ext_copp_ifout_find_entry (sem, ethertype);
+
+  if (!is_bind)
+    {
+      if (idx < 0)
+	return 0;
+      clib_memset (&sem->copp_ifout_entries[idx], 0,
```
yue-fred-gao, line 377: copp-ifout leaks table slots on unbind, and 16 bind/unbind cycles wedge CoPP.

   In copp_ifout_node.c, sonic_ext_copp_ifout_bind() with is_bind=0 memsets the entry but
   never decrements or reuses copp_ifout_n_entries. sonic_ext_copp_ifout_find_entry()
   skips entries with in_use == 0, so rebinding the same ethertype allocates a new slot.
   After SONIC_EXT_COPP_IFOUT_MAX_ENTRIES (16) trap install/remove cycles, which is normal
   on config reload or trap-group churn, binds fail with QUEUE_FULL and CoPP silently
   stops working. copp_ip2me_node.c already handles this correctly via
   sonic_ext_copp_ip2me_alloc_policer_slot(); mirror that here.

vppbld/plugins/sonic_ext/cli.c
```diff
@@ -147,6 +148,146 @@ VLIB_CLI_COMMAND (sonic_ext_ip2me_command, static) = {
...
+
+  while (unformat_check_input (input) != UNFORMAT_END_OF_INPUT)
+    {
+      if (unformat (input, "ethertype 0x%x", &ethertype))
+	;
+      else if (unformat (input, "ethertype %d", &ethertype))
+	;
+      else if (unformat (input, "policer %s", &policer_name))
+	;
+      else if (unformat (input, "match-ip4-ttl-expiring"))
```
yue-fred-gao: I think SAI_HOSTIF_TRAP_TYPE_TTL_ERROR in sonic is to punt ttl expired transit packets to control plane, which will send icmp with "ttl expired" code to the sender. This is how traceroute works. There is a few things.
1. TTL = 1 or 0 but destination is "for us" should not trigger the trap. It is common to set TTL 1 for packets intended to be single hop, like single hop ebgp.
2. copp_ifout_node is in the punt path. For transit packets with TTL error, they are not send to ip4-icmp-error node, which consumes the packet and sends icmp back. They are not punted. So the policer in copp_ifout_node is not applied to those packets.
3. the policer in copp_ifout_node incorrectly polices single hop for us packets, which is already policed by copp-ip2me. 

Is there sonic-mgmt to verify the ttl error policer? With vpp implementation, the policer won't affect packets punted to control plane because they are not punted. If you do want to apply the policer, it might be useful to control the rate that vpp sends icmp.

vppbld/plugins/sonic_ext/copp_ip2me_node.c
```diff
@@ -0,0 +1,754 @@
...
+      if (sem->copp_ip2me_addrs[i].in_use &&
+	  sem->copp_ip2me_addrs[i].addr == dst_addr)
+	return 1;
+    }
+  return 0;
+}
+
+/*
+ * Find the first in-use policer slot matching this packet: either the
+ * legacy shared IP2ME/SNMP/SSH address-match slot (dst_addr is one of
```
yue-fred-gao: what's the priority if a packet matches both a "match-addr" rule and "match-dport" rule?

vppbld/plugins/sonic_ext/copp_ip2me_node.c
```diff
@@ -0,0 +1,754 @@
...
+
+int
+sonic_ext_copp_ip2me_bind (const char *policer_name, int is_bind)
+{
+  return sonic_ext_copp_ip2me_bind_slot (policer_name, is_bind,
+					 SONIC_EXT_COPP_IP2ME_MATCH_ADDR, 0);
+}
+
+int
+sonic_ext_copp_ip2me_bind_bgp (const char *policer_name, int is_bind)
```
yue-fred-gao: Is the policer for all bgp packets? You need to use "src-port == 179 or dst-port == 179". It depends which side initiates the connection.

vppbld/plugins/sonic_ext/sonic_ext.api
```diff
@@ -72,3 +72,155 @@ autoreply define sonic_ext_mirror_encap_fixup_enable_disable
...
+    strictly by policer_name, so installing or removing the bgp trap
+    can never disturb sonic_ext_copp_ip2me_bind()'s separate IP2ME/
+    SNMP/SSH slot, or vice versa.
+
+    @param policer_name - name of an existing VPP policer object
+           (created via policer_add), looked up once via policer_main's
+           policer_index_by_name hash
+    @param is_bind - 1 to bind, 0 to unbind
+*/
+define sonic_ext_copp_ip2me_bind_bgp
```
yue-fred-gao: Can the API be more generic? instead of bind policer for bgp specifically, can it bind policer with certain condition? Is it over killed?

vppbld/plugins/sonic_ext/capture_node.c
nhegde-microsoft: Agreed. the opaque field is an optimization to avoid a linear search in the bound entries table. However, that is a small table of < 16 entries and is not worth taking up vnet space for. I will change this.
The classification cannot be done in `copp_ifout_node` though because this is a feature in the `interface-output` arc, which UDLD frames never reach on their own. `ethernet-input` forks sub-0x0600 frames to llc-input/snap-input before any TX interface (and hence that arc) is ever chosen.

vppbld/plugins/sonic_ext/sonic_ext.api
nhegde-microsoft: Sure, I can do that. Couple of clarifying questions:
1. By "feature" do you mean CoPP policing on the whole, or feature as in a VPP feature (this PR adds 3 VPP features on different arcs - `copp-ifout` on interface-output, `copp-ip2me` on ip4-punt, and `copp-ip2me-ip6` on ip6-punt)
2. Should/could this be a follow-up once [#288](https://github.com/sonic-net/sonic-platform-vpp/pull/288) merges since we will be extended its config struct?

vppbld/plugins/sonic_ext/copp_ip2me_node.c
nhegde-microsoft: Good point. Will fix.

vppbld/plugins/sonic_ext/copp_ifout_node.c
nhegde-microsoft: I missed this, thanks for catching it. WIll fix.

vppbld/plugins/sonic_ext/copp_udld_node.c
```diff
@@ -0,0 +1,299 @@
...
+#include <vnet/llc/llc.h>
+#include <vnet/snap/snap.h>
+#include <vnet/feature/feature.h>
+#include <plugins/linux-cp/lcp_interface.h>
+
+/* The ethertype/length value sonic-ext-copp-ifout's bind table uses
+ * for UDLD (see SwitchVppHostifTrap.cpp's buildClassifyMatchForTrapType,
+ * SAI_HOSTIF_TRAP_TYPE_UDLD case) -- both wire encodings this node
+ * handles carry this same value in the frame's 14th/15th bytes. */
+#define SONIC_EXT_COPP_UDLD_ETHERTYPE 0x0067
```
yue-fred-gao: 0x0067 is the length of the packet generated by PTF in the test. It is accidental and could change in the real deployment or if the test is modified.

vppbld/plugins/sonic_ext/copp_udld_node.c
```diff
@@ -0,0 +1,299 @@
...
+    clib_error_t *error = vlib_call_init_function (vm, llc_input_init);
+    if (error)
+      clib_error_report (error);
+  }
+
+  next_index = vlib_node_add_next (vm, llc_input_node.index, node_index);
+  lm->input_next_by_protocol[sap] = next_index;
+}
+
+#define SONIC_EXT_COPP_UDLD_LLC_SAP_PTF_TEST 0x30
```
yue-fred-gao: Is SAP 0x30 chosen by SONIC for UDLD? Using PTF_TEST in the name sounds like this is geared towards the test case specifically. If it is a value chosen by SONIC, we should remove PTF_TEST in the name. If it is a value chosen by the test, we need more generic solution.

vppbld/plugins/sonic_ext/capture_node.c
yue-fred-gao: We still need the udld capture node to redirect the packets to copp_ifout_node. If you can avoid taking the vnet space, we are good.

vppbld/plugins/sonic_ext/sonic_ext.api
yue-fred-gao: I think copp policing on the whole is fine. The feature arcs have inter-dependency that is hard to enable/disable on individually. You can use a new PR for the change. We also needs UTs once https://github.com/sonic-net/sonic-platform-vpp/pull/287 is merged. I plan to create issues for each change we have added but without test.

vppbld/plugins/sonic_ext/cli.c
nhegde-microsoft: You are right, but I think the solution needs some consideration. I don't think policing icmp responses is ideal because those are sent in response to multiple events (TTL, pmtu, etc.) and we don't want to police all of them together. We could police it in `ip4-input` where TTL policing is done, but that would mean patch of a core VPP file, which would be good to avoid, but we can do it if there is no other choice. Any suggestions on the right approach?

vppbld/plugins/sonic_ext/copp_ip2me_node.c
nhegde-microsoft: This is a little tricky because even when trap groups are distinct, the rules may overlap like you pointed out. I think it would make sense to match-dport first and then match-addr because match-dport represents a more specific trap (BGP) then the generic IP2ME/SSH catch all.

vppbld/plugins/sonic_ext/copp_ip2me_node.c
nhegde-microsoft: Agreed. Will fix.

vppbld/plugins/sonic_ext/sonic_ext.api
nhegde-microsoft: Making it completely generic seems like over kill in my opinion, at least for now when only BGP is the option. Let me make it generic by introducing a port option and renaming the API accordingly. If there are other use cases, the options can be expanded. Does this sound reasonable?

vppbld/plugins/sonic_ext/copp_udld_node.c
nhegde-microsoft: It was done that way initially and then I went with the separate UDLD node method making this length check unnecessary/fragile. I will remove it.

vppbld/plugins/sonic_ext/copp_udld_node.c
nhegde-microsoft: Will fix this. Identifying UDLD node based on previous comment will clean this up as well.

vppbld/plugins/sonic_ext/sonic_ext.api
yue-fred-gao: sounds good.

vppbld/plugins/sonic_ext/copp_ip2me_node.c
yue-fred-gao: that's what I thought too.

vppbld/plugins/sonic_ext/cli.c
nhegde-microsoft: Hi @yue-fred-gao, I gave it some more thought and this is my understanding and proposal. Let me know what you think.

Understanding: 
1. When no SAI trap is in place, TTL-expiring packets not destined locally are processed by `ip4-icmp-error`, which sends an ICMP reply back out the original ingress data interface (not the CPU-facing one).
2. When the SAI trap is in place, TTL-expired packets should be punted to the control plane. _**This is not implemented in VPP today.**_
3. With a CoPP policer in place, the punted traffic should be policed.
4. The CoPP tests today measure traffic received on the control (CPU-facing) side, not the ICMP responses sent.
5. The CoPP tests pass today because they use dest IP of "us", so the IP2ME node polices it — not TTL_ERROR enforcement.

Proposal (implement missing punt and police it):
- Add a small patch in `ip4-rewrite`, at the existing TTL-exceeded check, as follows:
  - If destined for us: skip all the below (already handled/policed via IP2ME).
  - If the TTL_ERROR SAI trap is in place:
    - If no policer resolved: punt to control plane, unpoliced.
    - If policer resolved: police; conforming packets go to control plane, violating packets are dropped.
  - If no SAI trap is in place: unchanged, existing behavior (`ip4-icmp-error`).
  
 NOTE that I prefer not to change the tests, so they will still pass for a different reason. But I will test this manually so the functionality will be in place.

yue-fred-gao: This is a good proposal. Is there a way to avoid patching ip4-rewrite? Is it possible to register a node before ip4-icmp-error to redirect the packet to the host? Does PR https://github.com/sonic-net/sonic-platform-vpp/pull/267 do something similar with the glean node?

PR #283, opened 2026-09-14
[vpp] Plugins and patches for Everflow

taran9999:
- Adds ACL PERMIT_MIRROR action (is_permit == 3)
- gre_tunnel_v3 api with per-tunnel outer TTL and GRE protocol override
- Adds egress_mirror_node.c to the sonic-ext plugin for deferred egress mirror path
- Enable gre_plugin.so plugin in startup conf

Required for implementing Everflow feature for enabling dataplane testing (issue: sonic-net/sonic-buildimage#25780 and sonic-net/sonic-buildimage#26047).

Supports changes in sonic-net/sonic-sairedis#2076


vppbld/patches/0019-acl-add-permit-mirror-action-for-everflow.patch
```diff
@@ -0,0 +1,313 @@
...
+ 
+diff --git a/src/plugins/acl/dataplane_node.c b/src/plugins/acl/dataplane_node.c
+index 471228726..bb71860f9 100644
+--- a/src/plugins/acl/dataplane_node.c
++++ b/src/plugins/acl/dataplane_node.c
+@@ -550,6 +550,102 @@ acl_fa_inner_node_fn (vlib_main_t * vm,
+ 	    next[0] = action ? next[0] : 0;
+ 	  }
+ 
++	  if (PREDICT_FALSE (action == 3))
```
nhegde-microsoft, line 189: can we use PERMIT_MIRROR instead of hardcoded value?

vppbld/patches/0019-acl-add-permit-mirror-action-for-everflow.patch
```diff
@@ -0,0 +1,313 @@
...
++		  vnet_sw_interface_is_valid (vnm, mirror_sw_if_index) &&
++		  vnet_sw_interface_is_admin_up (vnm, mirror_sw_if_index))
++		{
++		  int deferred = 0;
++
++		  /* A deferred rule matched on an ingress arc wants the clone
++		   * taken on the egress arc instead, so it reflects the
++		   * post-route/post-encap wire form. Hand it to the registered
++		   * platform stamper; clone here if there is none. */
++		  if ((mirror_flags & ACL_MIRROR_F_DEFERRED) && is_input &&
```
nhegde-microsoft, line 229: not sure I followed this. we are calling the stamper because we want an egressing packet mirrored, but the check for `is_input` indicates this is ingress processing. am I reading this incorrectly?

nhegde-microsoft:
does this need a vpp version bump in rules/vpp.mk?

aaronber0614:
Review notes (blocking):
1.  The new action == 3 block in 0019-acl-add-permit-mirror-action-for-everflow.patch collides with the existing ACL stateful-session path. acl_fa_track_session() already returns 3 for an established session, but that path skips ACL lookup and leaves match_acl_in_index and match_rule_index invalid. The new code then dereferences those indices as a mirror rule. Please distinguish a newly matched PERMIT_MIRROR rule from the established-session result, for example by requiring a completed ACL lookup and valid match indices.

2.  The immediate clone path always rewinds current_data to l2_hdr_offset. That is correct on an input ACL, but on an output ACL current_data already points to the egress L2 rewrite while l2_hdr_offset still describes the ingress frame. A tagged-ingress to untagged-egress packet therefore includes stale bytes before the egress Ethernet header in the mirrored copy. Please perform the L2 rewind only on the input path.

3.  sonic_ext_egress_mirror_enable_disable() increments or decrements active_egress_mirror_actions before sonic_ext_egress_mirror_arc_set() succeeds. If the transition fails, a retry can return success without retrying the arc transition. Please commit the refcount only after a successful transition or roll it back on failure.

Reviewed by AI agent on behalf of @aaronber0614.

aaronber0614:
> does this need a vpp version bump in rules/vpp.mk?

I believe there is a monthly release schedule when the version is incremented

aaronber0614 (changes requested):
Review notes (blocking):

The new action == 3 block in 0019-acl-add-permit-mirror-action-for-everflow.patch collides with the existing ACL stateful-session path. acl_fa_track_session() already returns 3 for an established session, but that path skips ACL lookup and leaves match_acl_in_index and match_rule_index invalid. The new code then dereferences those indices as a mirror rule. Please distinguish a newly matched PERMIT_MIRROR rule from the established-session result, for example by requiring a completed ACL lookup and valid match indices.

The immediate clone path always rewinds current_data to l2_hdr_offset. That is correct on an input ACL, but on an output ACL current_data already points to the egress L2 rewrite while l2_hdr_offset still describes the ingress frame. A tagged-ingress to untagged-egress packet therefore includes stale bytes before the egress Ethernet header in the mirrored copy. Please perform the L2 rewind only on the input path.

sonic_ext_egress_mirror_enable_disable() increments or decrements active_egress_mirror_actions before sonic_ext_egress_mirror_arc_set() succeeds. If the transition fails, a retry can return success without retrying the arc transition. Please commit the refcount only after a successful transition or roll it back on failure.

Reviewed by AI agent on behalf of @aaronber0614.

vppbld/patches/0020-gre-per-tunnel-ttl-and-protocol-override.patch
```diff
@@ -0,0 +1,429 @@
+From ca2347589d58e23735f1da463427e28b6d749486 Mon Sep 17 00:00:00 2001
```
yue-fred-gao, line 1: Work with AI to come up a different plan that can avoid this patch. 
1. Use TEB or ERSPAN GRE. SONiC everflow is closer to TEB except gre type, which you will override anyway.
2. Add a fixup node in ethernet-output arc to reset TTL and gre type. The node is attached to the selected GRE tunnel interface only to avoid impact to non-gre traffic.
3. Set VNET_BUFFER_F_LOCALLY_ORIGINATED in that node to avoid ip output node decrementing TTL
--------------------------------------------------------------------
CONTROL PLANE - sonic-sairedis, vslib/vpp/SwitchVppMirror.cpp
--------------------------------------------------------------------
1. tunnel.type = 2 (ERSPAN)  ->  tunnel.type = 1 (TEB)
2. Drop tunnel.session_id and the whole m_erspan_session_id_pool.
   The 1024-session cap disappears with it.
3. Drop tunnel.gre_protocol and tunnel.ttl from the tunnel create.
   Revert to the stock gre_tunnel_add_del_v2 API.
4. Drop the "session_ttl + 1" fudge.
5. Reject SAI_MIRROR_SESSION_ATTR_TTL == 0 at create/set instead of
   letting it black-hole every mirrored packet.
6. Add a small new VPP API call registering
   { mirror_tunnel_sw_if_index -> gre_protocol, ttl }
   with sonic_ext, and unregister it on session teardown.

Keep unchanged: pinErspanMonitor(), the monotonic m_next_gre_instance,
and the "pin before create" ordering.

--------------------------------------------------------------------
DATA PLANE - sonic_ext plugin
--------------------------------------------------------------------
New node "sonic-ext-mirror-encap-fixup" on the ethernet-output arc,
enabled only on the mirror tunnel's sw_if_index:

    ip4_header_t *ip = vlib_buffer_get_current (b);
    gre_header_t *gre = (gre_header_t *) (ip + 1);

    ip->ttl       = e->ttl;                 /* exact wire value */
    gre->protocol = clib_host_to_net_u16 (e->gre_protocol);
    ip->checksum  = ip4_header_checksum (ip);

    /* outer header built here, not forwarded: suppress underlay TTL-- */
    b->flags |= VNET_BUFFER_F_LOCALLY_ORIGINATED;

Plus: a per-interface config vec, an enable/disable API, and an
ip6-output twin if an IPv6 underlay is ever needed.

--------------------------------------------------------------------
WHY THIS PLACEMENT
--------------------------------------------------------------------
- adj-l2-midchain starts the ethernet-output arc AFTER prepending the
  outer header and running gre44_fixup (src/vnet/adj/adj_l2.c, around
  lines 120-128), with current_data already on the outer IPv4 header.
- The arc is per-interface, so there is zero cost on non-mirror traffic
  and no buffer cookie is needed.
- VNET_BUFFER_F_LOCALLY_ORIGINATED is clear at that point - verified
  that nothing in vnet/ethernet/, vnet/adj/, plugins/gre/, or
  vnet/interface_output.c sets it - and it is never cleared anywhere in
  the tree, so it survives to the underlay ip4-rewrite and suppresses
  the decrement.

--------------------------------------------------------------------
WHAT GETS DELETED
--------------------------------------------------------------------
- patch 0020 in full: 429 lines, including the gre_tunnel_v3 type and
  three new API messages
- m_erspan_session_id_pool
- the "session_ttl + 1" fudge

--------------------------------------------------------------------
BUGS FIXED AS A SIDE EFFECT
--------------------------------------------------------------------
- ttl = 255, the orchagent/SAI default, now emits 255 instead of 254.
- ttl = 0 becomes a config error instead of IP4_ERROR_TIME_EXPIRED on
  every mirrored packet.
- GRE_TUNNEL_TYPE_ERSPAN is no longer overloaded to mean "plain GRE",
  so the decap side and "show gre tunnel" stay honest.

--------------------------------------------------------------------
KNOWN COSTS
--------------------------------------------------------------------
- Loses gre_tunnel_dump visibility of the override. Add a
  "show sonic-ext mirror-encap" CLI instead.
- IPv6 underlay needs the second node instantiation. Unhandled today
  either way.
- DSCP/TOS still unset. Unchanged from today, but the node is now the
  natural place for it.

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
```diff
@@ -0,0 +1,123 @@
+From 473ca2c6d0880795ffc2accaab17b17bf1bd27d3 Mon Sep 17 00:00:00 2001
+From: taran9999 <tdwivedu@cisco.com>
+Date: Wed, 9 Sep 2026 13:50:59 -0700
+Subject: vnet: guard the drop node and L2 midchain against stale interface
+ state
+
+SONiC Everflow re-points a mirror session by deleting and recreating its
```
yue-fred-gao: This is a genuine problem in vpp. Please create an upstream PR in fd.io.
On the other hand, I am wondering how this is triggered. Traffic is redirected to the gre tunnel by ACL. If we remove ACL first then GRE tunnel, should that alleviate or even avoid the issue?

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
```diff
@@ -0,0 +1,123 @@
...
++/* During interface/tunnel teardown churn (e.g. mirror-clone packets whose
++ * source interface was just deleted) a buffer can reach the drop/punt node
++ * carrying an RX sw_if_index that no longer maps to a live interface. The
++ * per-interface counter, super-interface and drop-feature-arc lookups below
++ * index vectors/pools by that index without any bounds checking, so a stale
++ * index causes an out-of-bounds access and a worker SIGSEGV. Coalesce any
++ * invalid index to sw_if_index 0 (local0), which always exists, before it is
++ * used for a counter/pool/feature-arc lookup. */
++static_always_inline u32
++interface_drop_punt_safe_sw_if_index (vnet_main_t * vnm, u32 sw_if_index)
```
yue-fred-gao: Function vnet_get_sw_interface_or_null exists for the same purpose.

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
```diff
@@ -0,0 +1,123 @@
...
++      sw_if_index[0] = interface_drop_punt_safe_sw_if_index
++	(vnm, vnet_buffer (b[0])->sw_if_index[VLIB_RX]);
+ 
+       sw_if_index += 1;
+       n_left -= 1;
+@@ -1057,7 +1078,8 @@ interface_drop_punt (vlib_main_t * vm,
+ 	  {
+ 	    vlib_buffer_t *ob = bufs[orig_off + j];
+ 	    u32 orig = vnet_buffer2 (ob)->orig_rx_sw_if_index;
+-	    if (orig && orig != sw_if_index[0])
```
yue-fred-gao, line 118: We don't need interface_drop_punt_safe_sw_if_index and all the calls to it. It actually incorrectly increment counters on lo0, which has index 0. The real fix is below on orig_rx_sw_if_index, which may carry an invalid index. Akeel just changed the implementation in PR https://github.com/sonic-net/sonic-platform-vpp/pull/279. Please remove the change to interface_output.c and retest it.

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
```diff
@@ -0,0 +1,123 @@
...
+ 
+-		if (is_midchain)
++/*
++                 * A midchain adjacency can be torn down while packets are
++                 * still in flight for it, leaving fixup_func == NULL.
++                 * Calling it unconditionally jumps through a NULL pointer
++                 * and kills the worker. adj_midchain_fixup() in adj_dp.h
++                 * already performs this check; match it here.
++                 */
++                if (is_midchain && adj0->sub_type.midchain.fixup_func)
```
yue-fred-gao, line 54: Did you see a crash with backtrace showing adj0->sub_type.midchain.fixup_func===NULL? This doesn't look like coming from GRE tunnel recreation.

vppbld/patches/0020-gre-per-tunnel-ttl-and-protocol-override.patch
taran9999: Addressed in f255d8eb8c38e807cfbbac758dabef10ccdd9846

- Removed gre_tunnel_v3 patch and reverted to using gre_tunnel_v2 (also in sonic-net/sonic-sairedis@83544c907da86c8227fdea0cd28814f67cfc995d). m_erspan_session_id_pool also deleted.
- Added mirror_encap_fixup_node.c to sonic-ext (also added show sonic-ext mirror-encap to cli.c)
- Verified everflow/test_everflow_per_interface.py and everflow/test_everflow_testbed.py still fully pass.

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
taran9999: I did not see a crash with such backtrace. Deleted patch 21 in f255d8eb8c38e807cfbbac758dabef10ccdd9846 since this + interface_output.c hunks are no longer needed.

vppbld/patches/0021-vnet-guard-stale-interface-state-on-teardown.patch
taran9999: Patch was removed in f255d8eb8c38e807cfbbac758dabef10ccdd9846

taran9999:
@aaronber0614 , I've addressed your items in 76ac7c64dad073eac8ab557e2b201671b03ae149. Please review

aaronber0614 (approved):
LGTM. Please update description about removed gre_tunnel_v3 design

vppbld/patches/0020-adj-honour-ethernet-output-features-on-l2-adjacencies.patch
```diff
@@ -0,0 +1,68 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
yue-fred-gao, line 1: This patch fixes a bug in the code. Please create a PR in fd.io.

Also, copilot gives below comment. 

Main finding: adj-l2-rewrite is not a start node of the ethernet-output arc.

The ethernet-output arc declares exactly one start node in src/vnet/ethernet/init.c (lines 33-39): .start_nodes = VNET_FEATURES ("adj-l2-midchain").

But adj_l2_rewrite_inline() is shared by two nodes, and the arc-start branch is not guarded by is_midchain -- adj_l2_rewrite_node calls it with is_midchain = 0 and still dispatches the arc. Meanwhile the patch's adj_nbr_evaluate_feature hunk sets the flag for any VNET_LINK_ETHERNET adjacency, with no midchain distinction. Non-midchain ETHERNET adjacencies use adj-l2-rewrite per adj_get_rewrite_node() in src/vnet/adj/adj_internal.h (lines 49-50).

That combination is unsafe. add_next() in src/vnet/config.c (lines 30-51) computes the first feature's next index by iterating only cm->start_node_indices, so no edge from adj-l2-rewrite to the first feature node is ever created, and the index baked into the config string is adj-l2-midchain's. adj-l2-rewrite is registered with n_next_nodes = 1 in src/vnet/adj/adj_l2.c (lines 175-185) and, not being a start node, never gets its next vector grown. So a featured plain L2 adjacency would enqueue to an out-of-range next index -- assert in debug, silent misrouting or worse in release.

This is latent, not live: the Everflow GRE TEB adjacency is a midchain, so it goes through adj-l2-midchain, which is a start node and does get the edges. But the patch removes the thing that was accidentally protecting the other node, and anyone later enabling an ethernet-output feature on an interface carrying a plain L2 adjacency trips it.

I'd fold a third one-line hunk into the same patch, changing .start_nodes to VNET_FEATURES ("adj-l2-midchain", "adj-l2-rewrite"). That matches what adj_l2_rewrite_inline() already does unconditionally for both instantiations, and it's the fix upstream would want. The alternative -- restricting adj_nbr_evaluate_feature to midchain adjacencies -- is narrower but leaves the shared inline function lying about what it supports.

PR #284, opened 2026-09-15
[VPP] Exempt VXLAN and IP-in-IP tunnel decap from the spoofed-local/uRPF drop

dypet:
### Why I did it
A VTEP decapsulates tunnel traffic source-independently, but VPP's ip4-local/ip6-local run two source checks on every locally-terminated packet before dispatching to the decap node — spoofed-local (IP4_ERROR_SPOOFED_LOCAL_PACKETS) and uRPF/source-lookup-miss (IP4_ERROR_SRC_LOOKUP_MISS, and the ip6-local "unroutable host" check). A decap packet's outer source is arbitrary — it can be unroutable or coincide with one of the switch's own local addresses — so these checks silently drop legitimate VXLAN / IP-in-IP decap traffic before it reaches vxlan*-input / ipip*-input. On SONiC-VPP this makes outer-IPv4 VXLAN and IP-in-IP decap to a local VTEP fail. 

### How I did it
Added vppbld/patches/0019-ip-local-exempt-tunnel-decap-from-source-checks.patch, which exempts tunnel-decap candidates, UDP to the VXLAN dst port 4789 (RFC 7348) and IP-in-IP protocols 4/41, from those ip4-local/ip6-local source checks (single- and dual-loop paths, plus the IPv6 mirror). The decap node still gates real decap on the tunnel lookup, so a non-tunnel packet to a local address is still dropped there; anti-spoofing/uRPF is unchanged for every other protocol.

### How to verify it
Build a sonic-vpp image with this patch and run the VXLAN decap tests on t0-vpp (vxlan/test_vxlan_decap.py, vxlan/test_vnet_decap.py): the decap cases pass, and vppctl show errors shows the vxlan4-input/ipip4-input decap counters incrementing instead of SPOOFED_LOCAL_PACKETS / SRC_LOOKUP_MISS drops.


yue-fred-gao:
I am wondering if it is normal that "A decap packet's outer source is arbitrary — it can be unroutable or coincide with one of the switch's own local addresses". Is this a deficiency of a test or by design? Does it open a security hole if we allow that?

dypet:
> I am wondering if it is normal that "A decap packet's outer source is arbitrary — it can be unroutable or coincide with one of the switch's own local addresses". Is this a deficiency of a test or by design? Does it open a security hole if we allow that?

@yue-fred-gao Good point, I was working off of the current decap testcase implementation and this line from RFC 7348 §4.2: "The VTEP will use (*,G) joins. This is needed as the set of VXLAN tunnel sources is unknown and may change often, as the VMs come up / go down across different hosts.". But on closer inspection it looks like the testcase leaves ip_src unset in the sent packet and it inherits a default address which happens to also match the DUT’s Vlan1000 address; so I suspect it is not intentionally being set to a local address. I will rework this PR and test with the needed changes on the testcase side.

PR #287, opened 2026-09-17
Enable vpp test in ci pipeline

yue-fred-gao:
# [vpp] Run VPP's own test suite in CI; bump VPP pin; fix VXLAN decap-any multicast regression

## Why

The `vppbld/patches/` series has grown to 16 patches that modify core VPP
forwarding code — IPIP, VXLAN encap/decap, IP flow hash, ACL, sFlow and
`l2_bvi`. Until now nothing validated those patches against VPP's own
2600-case test suite. Breakage in patched code paths was only discoverable
downstream, in sonic-mgmt runs, long after the fact and with far worse
signal.

Turning the suite on found two real problems immediately:

1. **Patch 0017 broke VXLAN IPv6 multicast decapsulation.** The
   source-independent ("decap-any") lookup collides with the key space that
   multicast tunnels already occupy, so VPP decapsulated multicast frames
   from sources that have no tunnel instead of dropping them. This is a
   dataplane correctness bug, not a test artifact.

2. **The SONiC custom plugins were silently altering every test.** Only
   `sonic_ext` was `.default_disabled`; `ip_validate` and `tunterm_acl` were
   auto-loaded into every VPP instance. `ip_validate` additionally enables a
   dropping feature on *every* interface at creation, ahead of
   `ip4-lookup`/`ip6-lookup`, which produced unrelated failures across the
   suite.

3. Failures in about 100 tests that are known and skipped in bookworm. 
    The tests failed in trixie too but not skipped. We have to bump VPP pin to 
    pick up the fix of those failures.

Separately, the upstream has moved IPIP from `src/vnet/ipip/` to 
`src/plugins/ipip/`, so the affected patches had to be regenerated.

### 1. Bump the VPP pin and package version

- `vppbld/vpp_version`: `3f9e978d7` → `12154beed`
- `rules/vpp.mk`: `VPP_VERSION` `2606-0.6` → `2606-0.7`, so downstream
  sonic-buildimage builds do not pull stale debs predating this series.

### 2. Rebase patches onto the new pin

- `0009-ipip-add-mp2p-ipip-tunnel.patch` — retargeted from `src/vnet/ipip/`
  to `src/plugins/ipip/` following the upstream move.
- `0017-sonic-vxlan-vnet-source-independent-decap.patch` — context refresh
  plus the fix below.

Both were regenerated with `git diff` against a freshly applied tree rather
than hand-edited.

### 3. Fix the decap-any / multicast key-space collision (patch 0017)

VXLAN tunnels are keyed on the tunnel's *remote* address. Multicast tunnels
reuse that slot for the group address, and decap-any entries introduced by
0017 reuse it for the *local* address. For IPv6 those two overlap exactly:
a multicast frame whose source has no unicast tunnel missed the normal
lookup, fell into the decap-any probe, matched the multicast tunnel, and was
decapsulated instead of dropped.

`vxlan6_find_tunnel` now returns `decap_not_found` for a multicast outer
destination before consulting the decap-any entry — which is precisely the
upstream behaviour at that point. The same guard is applied to the IPv4 path;
it is not strictly required there (IPv4 stores multicast in the low half of
the key and decap-any in the high half) but it makes the invariant explicit
instead of leaving it resting on an unstated half-word distinction.

The registration side already refused to create decap-any entries for
multicast, so the decap side is now symmetric with it. No decap-any
functionality is lost.

Fixes `test_vxlan6.TestVxlan6.test_mcast_rcv` and
`test_vxlan6.TestVxlan6.test_mcast_rcv_custom_port`, which send 20 frames
from 20 sources of which only 10 have tunnels, and assert that exactly 10
are decapsulated.

### 4. Make the SONiC plugins opt-in

`.default_disabled = 1` added to `ip_validate` and `tunterm_acl` (and
already present on `sonic_ext`), matching how upstream treats `linux_cp`.
Both `docker-sonic-vpp` and `docker-syncd-vpp` `startup.conf.tmpl` already
enable all three explicitly, so **runtime behaviour is unchanged**.

Because the plugins no longer auto-load, their own suites now request them
explicitly via `extra_vpp_plugin_config`, the same mechanism the framework
uses for `unittest_plugin.so`.

### 5. Run `make test` in CI

New targets in `vppbld/Makefile`:

| target | purpose |
|---|---|
| `ci-build` | existing build, now stamped so a second CI step does not re-clone and rebuild |
| `ci-test-deps` | installs `tshark` (deliberately removed from `DEB_DEPENDS` by patch 0001) and re-asserts `install-dep`/`install-ext-deps` |
| `ci-test` | runs `make test` with `TEST_JOBS` and failed-log archiving |
| `ci-clean` | drops the stamp, artifacts and logs |

The stamp is invalidated by `vpp_version`, `patches/series`, any
`patches/*.patch` and any file under `plugins/`, so a changed patch or plugin
forces a rebuild.

`ci-test-deps` re-runs `install-dep` unconditionally because apt state is
container state while the stamp lives on a persistent workspace mount — a
stamp written by an earlier container would otherwise leave the toolchain
missing.

Failed-test logs are archived by setting `WORKSPACE` together with
`COMPRESS_FAILED_TEST_LOGS=yes`. VPP's `FAILED_DIR` contains only symlinks
into `/tmp/vpp-unittest-*`, which an artifact upload cannot follow;
`compress_failed.sh` gzips them into real files under `$WORKSPACE/archives`,
and it runs inside VPP's own `test` recipe before the failure propagates.

`.azure-pipelines/build-template.yml` gains a "Run VPP make test" step and
publishes `vpp-test-logs-<debian_version>` on failure.

## How to verify
Check the PR checker CI result
Build and run the full suite:

    NOBULLSEYE=1 NOBUSTER=1 NOBOOKWORM=1 make sonic-slave-bash
    cd platform/vpp/vppbld
    make ci-build; make ci-test

Run only the tests fixed here:

    NOBULLSEYE=1 NOBUSTER=1 NOBOOKWORM=1 make sonic-slave-bash
    cd platform/vpp/vppbld/repo
    make test PLATFORM=vpp TEST=vxlan6.TestVxlan6.test_mcast_rcv*

Confirm the plugins are no longer auto-loaded, and that SONiC still enables
them:

    grep -A4 VLIB_PLUGIN_REGISTER platform/vpp/vppbld/plugins/*/*.c
    grep -n 'plugin .*_plugin.so' platform/vpp/docker-*-vpp/conf/startup.conf.tmpl


aaronber0614 (changes requested):
Review note (blocking):

The VPP test-suite integration and VXLAN multicast fix look sound, but I found two blocking issues.

1. vppbld/patches/series no longer has a trailing newline. The repo_clone loop in vppbld/Makefile uses plain while read, so it silently stops at patch 0017 and never applies 0018-acl-match-on-ingress-interface.patch. Current sonic-sairedis writes vl_api_acl_rule_t.in_sw_if_index, which is introduced by 0018, so the resulting VPP headers break the downstream libsaivpp build. Please restore the newline; making the loop EOF-safe would also prevent this class of failure.

2. The new VPP pin includes upstream's IPIP module-to-plugin migration, but the required integration is incomplete. Current sonic-sairedis still includes <vnet/ipip/ipip.api*.h>, while those headers are now installed under <vpp_plugins/ipip/...>. In addition, the SONiC VPP startup profiles disable unlisted plugins and currently omit ipip_plugin.so; sonic-sairedis unconditionally resolves the IPIP API and asserts that it exists during initialization. Please land the companion sonic-sairedis include update and enable ipip_plugin.so in both startup templates and both generated-config defaults before consuming this pin.

Reviewed by AI agent on behalf of @aaronber0614.

yue-fred-gao:
> Review note (blocking):
> 
> The VPP test-suite integration and VXLAN multicast fix look sound, but I found two blocking issues.
> 
> 1. vppbld/patches/series no longer has a trailing newline. The repo_clone loop in vppbld/Makefile uses plain while read, so it silently stops at patch 0017 and never applies 0018-acl-match-on-ingress-interface.patch. Current sonic-sairedis writes vl_api_acl_rule_t.in_sw_if_index, which is introduced by 0018, so the resulting VPP headers break the downstream libsaivpp build. Please restore the newline; making the loop EOF-safe would also prevent this class of failure.
> 2. The new VPP pin includes upstream's IPIP module-to-plugin migration, but the required integration is incomplete. Current sonic-sairedis still includes <vnet/ipip/ipip.api*.h>, while those headers are now installed under <vpp_plugins/ipip/...>. In addition, the SONiC VPP startup profiles disable unlisted plugins and currently omit ipip_plugin.so; sonic-sairedis unconditionally resolves the IPIP API and asserts that it exists during initialization. Please land the companion sonic-sairedis include update and enable ipip_plugin.so in both startup templates and both generated-config defaults before consuming this pin.
> 
> Reviewed by AI agent on behalf of @aaronber0614.

Good points. Will fix those. Thanks!

yue-fred-gao:
created PR https://github.com/sonic-net/sonic-sairedis/pull/2082

yue-fred-gao:
Created PR https://github.com/sonic-net/sonic-sairedis/pull/2082 to adapt to the ipip moving to plugin. sonic-buildimage PR https://github.com/sonic-net/sonic-buildimage/pull/27039 runs build with both PRs and it passes all the tests.

dypet (approved):
LGTM

aaronber0614 (approved):
LGTM

Issue #285, opened 2026-09-15
Disable/enable sonic-ext plugin features via config file


yue-fred-gao:
sonic-ext plugin has many features. They are for sonic compliant and each is too small to be an individual plugin. We can't disable a feature by not enabling the plugin because that will disable other features. sonic-ext plugin can add api for enabling a feature but that requires vpp sai code to call the api based on some runtime information, such as platform hwsku. A better way is enabling/disabling a feature through config file, such as startup.conf or startup config file. The latter contains initial config in vppctl clis to be executed after vpp is started. Each platform can have its own startup.conf or startup config. This decouples feature enabling from vpp sai.

PR #288, opened 2026-09-21
Select sonic_ext plugin features from config (Phase 1)

AkeelAli:
## Select sonic_ext plugin features from config (Phase 1)

Addresses #285.

The `sonic_ext` plugin bundles several small SONiC-compliance behaviours into one
shared object. The only granularity available today is loading or not loading
`sonic_ext_plugin.so` — all or nothing.

This PR adds per-feature selection driven from
a config file.

### The stanza

Three keywords, each accepting `on|enable|off|disable`, all defaulting to **on**, so
an untouched tree behaves exactly as before:

```
sonic-ext {
  punt-via-member   off
  host-xc           on
  drop-member-stats off
}
```

An unrecognised keyword returns a `clib_error_t`, so VPP fails at boot rather than
running an unintended configuration.

`capture` deliberately has no keyword. It produces the per-buffer cookie the other
features read, so it is **derived**: attached iff `punt-via-member || drop-member-stats`.

`punt-via-member` is one behaviour spanning three nodes — `capture`,
`aggr-tap-redirect` and `glean-redirect` — which is why they share a keyword. That
grouping is pre-existing; all three already tested the same `sem->punt_via_member`
flag, and the existing `vppctl sonic-ext punt-via-member on|off` already grouped them.

### A disabled feature is never attached

The point of doing this at boot rather than from the CLI: a feature turned off in
`startup.conf` is **never added to its feature arc**, rather than attached and
short-circuiting per packet. That reclaims the dispatch cost entirely.

This required splitting init into three steps:

| Phase | Function | Does |
|---|---|---|
| init | `sonic_ext_init` | registers the LCP vft, seeds default flags, **wires nothing** |
| config | `sonic_ext_config` | records the stanza; assignment only |
| apply | `sonic_ext_apply_config` | wires only what survived, setting the latches |

Apply has to be a `VLIB_MAIN_LOOP_ENTER_FUNCTION` rather than part of the config
function, because a config function only runs when its stanza is *present* — the
defaults would otherwise never be applied. At that point no LCP pair exists yet
(SAI creates them once the main loop is running), so it is the latches, not the
walks, that drive per-interface attachment later.

Note the CLI is unchanged and still only *gates* — it clears the flag but leaves the
node on its arc. 

### Delivery

Two paths, both landing in the same generated config:

| | Scope | Cost |
|---|---|---|
| stanza in `startup.conf.tmpl` | every image built from the tree | rebuild |
| `SONIC_EXT_CONFIG=host-xc=off` in `/etc/sonic/vpp/syncd_vpp_env` | one switch | restart syncd |

`vpp_init.sh` copies the template first and appends the `$SONIC_EXT_CONFIG` stanza
last, so the generated file can hold two `sonic-ext { }` blocks. They merge
**per keyword**, environment winning:

```
`sonic_ext_init()` defaults (all enabled)  <  startup.conf.tmpl  <  $SONIC_EXT_CONFIG
```

### Latent bug fixed

`aggr-tap-redirect` was attached unconditionally in `sonic_ext_lcp_pair_add_cb` —
the only attach path not behind a latch. Without the new `aggr_tap_redirect_enabled`
latch, `punt-via-member off` would still have wired it onto every aggregate host tap.

### Validation

On t1-lag-vpp (32 ports, 8 bonds):

- all-off boot — zero `sonic-ext-*` features on any interface
- `punt-via-member off` + `drop-member-stats on` — capture correctly still attached,
  glean and aggr-tap-redirect absent
- runtime re-enable — nodes attach, including on pairs that already existed; a second
  `on` is idempotent
- template + environment together — per-keyword merge confirmed

`show sonic-ext` now also reports `drop-member-stats` and `capture (derived)`. Note it
reports the *toggle*, not arc membership — `show interface features <if>` is the
authoritative check, since a boot-time `off` and a CLI `off` both read `off` but only
the first is actually detached.

### Out of scope

No change to `sonic-sairedis`. `ip2me`, `l2-trap-fixup` and `l2-vlan-filter` are
entered from saivpp, not wired by the plugin, so `startup.conf` cannot reach them.

### Phase 2

To follow in separate PRs:

- **The remaining three features** — `ip2me`, `l2-trap-fixup`, `l2-vlan-filter`.
  These need a different mechanism because saivpp, not the plugin, decides whether to
  attach them. The leading candidate is a `sonic_ext_feature_get` binary API: keep all
  seven keywords in the one stanza and have saivpp *query* before deciding, rather than
  VPP refusing after the fact. The alternative is reading per-HWSKU keys from
  `sai_vpp.profile`, which syncd already loads.
- Possibly a `drop-member-stats` CLI toggle, for symmetry with the other two knobs.
- **HLD documentation** — if needed: design, feature inventory, classification and the full test
  matrix.


AkeelAli:
All sanities pass: https://github.com/sonic-net/sonic-buildimage/pull/29703. Phase 1 can be reviewed/merged.

AkeelAli:
Replaced with full Phase 1 & 2 commit. Please see: https://github.com/sonic-net/sonic-platform-vpp/pull/291

